PR #21, opened 2025-07-07
Refactor VOD Domain Module

### LINKED ISSUE #20: create VOD Domain Module

mostafa-n3ma:
✅ What was done:
•	Merged the previously separated domain modules:
•	actors, movies, series, and search
•	→ into a single unified module: vod_domain
•	Standardized domain structure with:
•	model/ → Core models: Movie, TvSeries, Actor, etc.
•	repository/ → Clean repository interfaces
•	usecase/ → Focused use cases for search and history

🧹 Cleanup:
•	Removed unused domain modules from settings.gradle.kts and build.gradle.kts
•	Renamed domain root folder from Domains → domains for consistency
•	Ensured vod_domain is a pure Kotlin module (no Android dependencies)

🔁 Model Enhancements:
•	Replaced Long timestamp with LocalDateTime using kotlinx.datetime
•	Defined MediaFilters and SearchHistoryItem for scalable search logic
•	Ensured all models are reusable across multiple features (not just search)

💡 Why this change?

This refactor:
•	Centralizes all video-on-demand logic under one cohesive domain
•	Improves maintainability and reusability across features
•	Reduces duplication and confusion from overly granular domain modules

if you have any notes please let me know ....


domain/vod/src/main/java/model/MediaFilters.kt
```diff
@@ -0,0 +1,7 @@
+package model
+
+data class MediaFilters(
+    val releaseYear: Int? = null,
+    val genres: List<String>? = null,
```
haideraljawasim: The genres should be enum class , not strings

domain/vod/src/main/java/model/MediaFilters.kt
```diff
@@ -0,0 +1,7 @@
...
+    val imdbRating: Float? = null
```
haideraljawasim: I think we should use Double here instead of Float

domain/vod/src/main/java/model/SearchHistoryItem.kt
```diff
@@ -0,0 +1,11 @@
+package model
+
+import kotlinx.datetime.LocalDateTime
+
+data class SearchHistoryItem(
+    val query: String,
+    val category: SearchCategory,
+    val filters: MediaFilters? = null,
+    val language: String,
```
haideraljawasim: here also i think it should be enum class for language mapping will be easier and safer

domain/vod/src/main/java/model/TvSeries.kt
```diff
@@ -0,0 +1,14 @@
+package model
+import kotlinx.datetime.LocalDate
+
+data class TvSeries(
+    val id: String,
+    val imageUrl: String,
+    val title: String,
+    val description: String,
+    val seasons: List<String>,
```
haideraljawasim: why would you used here a list of string why didn't you create data class represent season with list episodes for each season or maybe a later get them by series id  which i think is the best approach , cause actually don't get it why list of string

domain/vod/src/main/java/model/TvSeries.kt
```diff
@@ -0,0 +1,14 @@
...
+    val actors: List<String>,
+    val releaseDate: LocalDate,
+    val genres: List<String>,
```
haideraljawasim: same genres should be enum here

domain/vod/src/main/java/model/TvSeries.kt
```diff
@@ -0,0 +1,14 @@
+package model
+import kotlinx.datetime.LocalDate
+
+data class TvSeries(
+    val id: String,
+    val imageUrl: String,
+    val title: String,
+    val description: String,
+    val seasons: List<String>,
+    val actors: List<String>,
```
haideraljawasim: Same why list of strings why not List of Actors ID's or may we get actors later by movie id  since if you entered a movie and then clicked the actors you would not see just it name you will see its image and charter add to the name , so we are literally will not get any thing from this list

domain/vod/src/main/java/model/movie.kt
```diff
@@ -0,0 +1,13 @@
+package model
+import kotlinx.datetime.LocalDate
+
+data class Movie(
```
haideraljawasim: It miss movie duration here

domain/vod/src/main/java/model/movie.kt
```diff
@@ -0,0 +1,13 @@
...
+    val id: String,
+    val imageUrl: String,
```
haideraljawasim: i suggest posterImageUrl the name will be more clear

domain/vod/src/main/java/model/movie.kt
```diff
@@ -0,0 +1,13 @@
...
+    val title: String,
+    val description: String,
+    val actors: List<String>,
```
haideraljawasim: same i don't get why list of string we are literally would not get any thing useful from this list

domain/vod/src/main/java/model/movie.kt
haideraljawasim: also it miss if the movie saved or not the user should be able to save movie , we would need this later also since this a shared entity i think we should add it now

domain/vod/src/main/java/model/SearchCategory.kt
```diff
@@ -0,0 +1,3 @@
+package model
+
+enum class SearchCategory { MOVIE, TV_SERIES, ACTOR }
```
haideraljawasim: Not necessary comment but would be more readable if like each value with new line

domain/vod/src/main/java/model/Actor.kt
```diff
@@ -0,0 +1,10 @@
+package model
```
haideraljawasim: we have been told a lot in public code reviews to not call our entities model so i think you should rename this package to entities

domain/vod/src/main/java/model/Actor.kt
```diff
@@ -0,0 +1,10 @@
...
+
+data class Actor(
+    val id: String,
+    val imageUrl: String,
+    val name: String,
+    val age: Int?,
+    val region: String?,
+    val lastShow: String?
```
haideraljawasim: These values we don't need them also Im not sure if we should keep or remove them but i suggest removing them since we may can't get this info from api

domain/vod/src/main/java/model/movie.kt
haideraljawasim: also you should add trailer url

haideraljawasim (changes requested):
Well done Mustafa you may need small refactoring ,
Last thing i think you should add some exceptions in this PR if we gonna work on interface implementation  or ui today later

domain/vod/src/main/java/model/Actor.kt
AhmadKharfan: @haideraljawasim 
Our domain model should reflect what an Actor means to the product, not what the current API provides.  
The  region and lastShow are nullable for now if future features never use them, we can safely drop them later.

domain/vod/src/main/java/model/MediaFilters.kt
```diff
@@ -0,0 +1,7 @@
+package model
+
+data class MediaFilters(
```
AhmadKharfan: MediaFilters looks like a UI-state holder (user’s current filter choices), not a domain object or VOD domain.
Unless a use-case needs it, move it to the presentation layer (UiState) and keep the domain layer for business entities only.

domain/vod/src/main/java/model/SearchHistoryItem.kt
```diff
@@ -0,0 +1,11 @@
+package model
+
+import kotlinx.datetime.LocalDateTime
+
+data class SearchHistoryItem(
```
AhmadKharfan: SearchHistoryItem is screen-level data (just what the user typed + filters) and isn’t reused by core VOD business rules, so it belongs in UI state / presentation, not in the vod domain module.

domain/vod/src/main/java/repository/ActorRepository.kt
```diff
@@ -0,0 +1,7 @@
+package repository
+
+import model.Actor
+
+interface ActorRepository {
+    suspend fun searchActors(query: String, language: String): List<Actor>
```
AhmadKharfan: this searchActors() method belong in a SearchRepository within the search domain, not in the core Actor repository.

domain/vod/src/main/java/repository/MovieRepository.kt
```diff
@@ -0,0 +1,9 @@
+package repository
+
+import model.MediaFilters
+import model.Movie
+
+interface MovieRepository {
+    suspend fun searchMovies(query: String, filters: MediaFilters?, language: String): List<Movie>
```
AhmadKharfan: Same for searchMovies() method it should be in SearchRepository within the search domain.

domain/vod/src/main/java/repository/TvSeriesRepository.kt
```diff
@@ -0,0 +1,8 @@
+package repository
+
+import model.MediaFilters
+import model.TvSeries
+
+interface TvSeriesRepository {
+    suspend fun searchTvSeries(query: String, filters: MediaFilters?, language: String): List<TvSeries>
```
AhmadKharfan: Also here this method belong to  SearchRepository.

domain/vod/src/main/java/model/MediaFilters.kt
mostafa-n3ma: All the genres would be like 20 genres I don't think it's right to create enum for that

domain/vod/src/main/java/model/MediaFilters.kt
mostafa-n3ma: The imdb rating is from 0 to 99.9 it won't be more then that so I think float is enough

domain/vod/src/main/java/model/SearchHistoryItem.kt
mostafa-n3ma: Agree that would be better 👍🏻

domain/vod/src/main/java/model/TvSeries.kt
mostafa-n3ma: Agree that would be better nice catch 👍🏻

domain/vod/src/main/java/model/TvSeries.kt
mostafa-n3ma: Let me check on that first

domain/vod/src/main/java/model/movie.kt
mostafa-n3ma: We can save content 🤨!

domain/vod/src/main/java/model/Actor.kt
mostafa-n3ma: I believe it is the opposite no entities 😅 they are just models

domain/vod/src/main/java/model/MediaFilters.kt
mostafa-n3ma: Media filters will be saved as properties in the searchHistoryItem so we can get the same filtered data or make new api cal when the filters change for the same query

domain/vod/src/main/java/model/SearchHistoryItem.kt
mostafa-n3ma: We need to save history locally in data base 
Also call this in the CRUD useCases

domain/vod/src/main/java/repository/MovieRepository.kt
mostafa-n3ma: Same replay please check my replay on the past comment

mostafa-n3ma:
Thank you guys for the valuable notes I will work on them immediately 👍🏻👍🏻👍🏻

domain/vod/src/main/java/model/MediaFilters.kt
mostafa-n3ma: sorry it is a good idea 
done ✅

domain/vod/src/main/java/model/MediaFilters.kt
mostafa-n3ma: IMDB ratings are limited in range and precision, and we’re not doing complex calculations here, so Float keeps the model lightweight. If we later see any need for more precision, we can refactor

domain/vod/src/main/java/model/SearchHistoryItem.kt
mostafa-n3ma: done ✅

domain/vod/src/main/java/model/TvSeries.kt
mostafa-n3ma: done ✅

domain/vod/src/main/java/model/movie.kt
mostafa-n3ma: i though of going with list of actors since we are going to use this in detailes screen two but this model would become heavy
mybe list of actors id's is better

domain/vod/src/main/java/model/Actor.kt
ZainabAlnajjar: @mostafa-n3ma they are entities for the domain layer, 
the mentors made it clear in many code reviews, so please rename it
(even yesterday if you could remember when Bareq said: "no need to add the suffix entity to the data classes like MovieEntity or ActorEntity, but name it directly as Movie or Actor would be enough cuz they are in the entity package")

domain/vod/src/main/java/model/SearchHistoryItem.kt
mostafa-n3ma: Maybe we need to discuss it today in the meeting

domain/vod/src/main/java/repository/ActorRepository.kt
mostafa-n3ma: I agree that searchActors() is more about search logic, not actor management itself.
I’ll move searchActors() to a SearchRepository and keep ActorRepository focused on actor-specific actions like addToFavorites() and getActorDetails()

ZainabAlnajjar: @AhmadKharfan just to clarify on what Mostafa said, there is no such thing as search domain or search repository as we discussed yesterday

searching is a use case that can deal with many repos, if you search for actors? by actor repo, search for movie? by movie repo, and so on

mostafa-n3ma:
dear @haideraljawasim  and @AhmadKharfan 
Thanks for the valuable review; it helped improve the code 
Please check the latest updates

domain/vod/src/main/java/model/Actor.kt
mostafa-n3ma: Got it rename the packadge to entity @ZainabAlnajjar

domain/vod/src/main/java/repository/ActorRepository.kt
mostafa-n3ma: I was wrong @ZainabAlnajjar this is not actor CRUD Logic we are not adding or deleting any actor data we are just searching so in the search repository implementation we can use the actor data source and retrive data 
so search repository is clear

ZainabAlnajjar: he deleted the old reply haha

domain/vod/src/main/java/repository/SearchHistoryRepository.kt
```diff
@@ -0,0 +1,9 @@
+package repository
+
+import model.SearchHistoryItem
+
+interface SearchHistoryRepository {
+    suspend fun getSearchHistory(): List<SearchHistoryItem>
```
ElhamFadel:
```suggestion
    suspend fun getSearchHistory(): Flow<List<SearchHistoryItem>>
```

ElhamFadel (changes requested):

domain/vod/src/main/java/model/MediaFilters.kt
AhmadKharfan: I see but searchHistoryItem itself should live in UI state it only stores UI context, and the domain layer should stay business only

domain/vod/src/main/java/model/movie.kt
haideraljawasim: we can save content to watch later you can check the design  i mean not to download but like add in saved list  
Screenshot 2025-07-08 at 1 23 10 PM
from seach screen where the icon on the image represent if the movie or series  saved or not

domain/vod/src/main/java/model/SearchHistoryItem.kt
AhmadKharfan: We can keep SearchHistoryItem as a UI-state object and let each CRUD use-case expose its own Params (or nested data class) the presentation layer does the mapping, then hands the domain an inner data class that the use-case owns

```
// ---------- presentation (ViewModel) ----------
val uiState = SearchHistoryItem(query = "Matrix", filters = genreFilters)

// Map UI state ➜ use-case params
val params = SearchMoviesUseCase.Params(
    query   = uiState.query,
    filters = uiState.filters
)

searchMoviesUseCase(params)


// ---------- domain ----------
class SearchMoviesUseCase(
    private val repo: SearchRepository
) {

    // Inner data class visible only to the use-case
    data class Params(
        val query: String,
        val filters: List<Filter>
    )

    operator fun invoke(params: Params) =
        repo.search(params.query, params.filters)
}
```

So:

SearchHistoryItem stays in the UI state.

The ViewModel converts it to Params (the inner data class).

The domain layer receives only its own primitives and never depends on presentation or data storage.

AhmadKharfan: Let’s talk and discuss more in the daily

domain/vod/src/main/java/repository/ActorRepository.kt
AhmadKharfan: @ZainabAlnajjar We discussed that search isn’t its own business domain, but it’s fine to keep a repository for the search logic.

AhmadKharfan (commented):
Great job, Mostafa your code looks great!
We can talk more about the SearchHistoryItem and filters in the daily.
Aashet eidak :)

domain/vod/src/main/java/repository/SearchHistoryRepository.kt
AhmadKharfan: The Figma design doesn’t show live-updating search, so a simple `suspend` fetch is enough lighter and avoids the extra overhead of streaming data with `Flow`

domain/vod/src/main/java/model/Language.kt
```diff
@@ -0,0 +1,11 @@
+package model
+
+enum class Language(val code: String) {
+    ENGLISH("en"),
+    ARABIC("ar");
+
+    companion object {
+        fun fromCode(code: String): Language? =
+            Language.entries.find { it.code == code }
+    }
```
devfalah: There's no need for this if you need to map data to a domain, you can create a mapper

domain/vod/src/main/java/usecase/SearchActorsUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+    private val historyRepo: SearchHistoryRepository
+) {
+    suspend fun execute(query: String, language: Language): List<Actor> {
+        historyRepo.addSearchHistoryItem(
+            SearchHistoryItem(
+                query,
+                SearchCategory.ACTOR,
+                null,
+                language,
+                Clock.System.now().toLocalDateTime(TimeZone.currentSystemDefault())
```
devfalah: Create a reusable function that gets the current date time.

```kotlin
fun LocalDateTime.Companion.now(
    timeZone: TimeZone = TimeZone.currentSystemDefault()
) = Clock.System.now().toLocalDateTime(timeZone)

//Usage:
LocalDateTime.now()
```

domain/vod/src/main/java/usecase/SearchMoviesUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+    private val historyRepo: SearchHistoryRepository
+) {
+    suspend fun execute(query: String, filters: MediaFilters?, language: Language): List<Movie> {
+        historyRepo.addSearchHistoryItem(
+            SearchHistoryItem(
+                query,
+                SearchCategory.MOVIE,
+                filters,
+                language,
+                Clock.System.now().toLocalDateTime(TimeZone.currentSystemDefault())
```
devfalah: same

domain/vod/src/main/java/repository/SearchHistoryRepository.kt
mostafa-n3ma: just like Ahmed says 👌🏻

domain/vod/src/main/java/model/Genre.kt
```diff
@@ -0,0 +1,15 @@
+package model
+
+enum class Genre {
```
AimanYosofiUlt: I disagree with putting the Genre as an enum class, Genres can change over time, and there are so many Genres
I can argue we should have the ANIME genre here

AimanYosofiUlt (changes requested):

domain/vod/src/main/java/usecase/AddRecentWatchingUseCase.kt
```diff
@@ -0,0 +1,10 @@
+package usecase
+
+import repository.SearchHistoryRepository
+import usecase.search.RecentWatchingItem
+
+class AddRecentWatchingUseCase(
```
haideraljawasim: Rename recent watching to recent viewed as the designer make it clear this part for recent viewed

domain/vod/src/main/java/usecase/RemoveSearchHistoryItemUseCase.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import repository.SearchHistoryRepository
+import usecase.search.SearchHistoryItem
+
+class RemoveSearchHistoryItemUseCase(
+    private val historyRepo: SearchHistoryRepository
+) {
+    suspend fun execute(item: SearchHistoryItem) {
+        historyRepo.removeSearchHistoryItem(item)
+    }
```
haideraljawasim: I'm not get it here why we delete by object why not by id

domain/vod/src/main/java/usecase/search/MediaFilters.kt
```diff
@@ -0,0 +1,7 @@
+package usecase.search
+
+data class MediaFilters(
+    val releaseYear: Int? = null,
+    val genres: List<String>? = null,
```
haideraljawasim: genres we already switched to use enum class for them but here still list of string why ?

domain/vod/src/main/java/usecase/RemoveSearchHistoryItemUseCase.kt
mostafa-n3ma: As haider says 
It is already checking item by id under the hood 
So by id is straight forward

AhmadKharfan: Actually i thought about deleting by id
but i don’t see an id inside SearchHistoryItem since we’re passing the whole object.
i’m thinking of updating the data class to include an id, but wanted to discuss it with you first.
anyway i’ll go ahead and add the id to the data class and pass it through the use-case

domain/vod/src/main/java/usecase/search/MediaFilters.kt
AhmadKharfan: We already switched it in the TvSeries entity, but we forgot to update MediaFilters I’ll fix that 😅

domain/vod/src/main/java/usecase/search/MediaFilters.kt
```diff
@@ -0,0 +1,9 @@
+package usecase.search
+
+import entity.Genre
+
+data class MediaFilters(
+    val releaseYear: Int? = null,
```
haideraljawasim: I checked the design again now and noticed the year filter should be range and not one release year so we should refactor this

domain/vod/src/main/java/model/Genre.kt
ZainabAlnajjar: as Aiman said, enum is not the best way to handle genres, there could be more or less types in the future, in that case we will need to edit on all layers just to add the new genre and we should specify every think in static way

I am suggesting doing something like the default categories in Tudee, like the following:

we will have a genre class that has the (id, name, imgPath), the image would come from asset file and will be mapped in the repository

in this case we can search by the genre id in filters and the localization of the name would come from the api itself
another advantage is that even if we changed the Api later the other layers won't get effected, everything would be handled from the DataSource and mapper without the other layers, another thing is that the genre of the movies has different names and ids than the. so making a data class would be more flexible

domain/vod/src/main/java/entity/Actor.kt
```diff
@@ -0,0 +1,10 @@
+package entity
+
+data class Actor(
+    val id: String,
```
ZainabAlnajjar: why the id is string? it should be long, same for the other entities (even the API returns number)

domain/vod/src/main/java/model/Genre.kt
haideraljawasim: What do you mean by “we will need to edit all layers”?

Adding a new Genre as a data class will not affect  the reset ? like how so 
also adding new genre to enum class will not effect the whole logic of layers it will just effect the mapping between them like just add one line to the When statement to get the new Genre and i think this completely normal 
If you meant just returning the genre name as a String from the data source  that’s also not correct, because that would make the domain layer depend on the data and break encapsulation
the domain should have its own expected structure and not rely on the internal format of the data.

@ZainabAlnajjar

ZainabAlnajjar: @haideraljawasim the domain shouldn't get changed when a new data came along
the enum class would be useful for a static values like task status, but genres could change over time

also, the api have the localization name so that would be another advantage

haideraljawasim: suppose the api return "blah" but ment to  be hmmm let say Action  Genre and our domain required from us to filter by specified genre 
what to do here in this case ?
do we display the genre as returned from api ?

NOTE the return value from API change by time or data source

haideraljawasim: we defiantly will need to map to something that we can use and display as we want

domain/vod/src/main/java/entity/Actor.kt
```diff
@@ -0,0 +1,10 @@
+package entity
+
+data class Actor(
```
ZainabAlnajjar, line 3: we also need the gender of actor, maybe add it as isMale boolean

domain/vod/src/main/java/entity/Genre.kt
```diff
@@ -0,0 +1,28 @@
+package entity
+
+enum class Genre {
```
ZainabAlnajjar, line 3: missing Science Fiction

ZainabAlnajjar: also `Action & Adventure`

ZainabAlnajjar: SCI_FI -> Sci-Fi & Fantasy

domain/vod/src/main/java/usecase/search/SearchHistoryOutputItem.kt
```diff
@@ -0,0 +1,8 @@
+package usecase.search
+
+data class SearchHistoryOutputItem(
+    val id: Long? = null,
+    val title: String,
+    val posterImageUrl: String,
```
ZainabAlnajjar: actor doesn't have poster image; it would be better to just name it `imageUrl`

ZainabAlnajjar: or it would be better to make another output for the actor cuz it also don't have the isSaved parameter

domain/vod/src/main/java/usecase/SearchTvSeriesUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+
+class SearchTvSeriesUseCase(
+    private val searchRepository: SearchRepository,
+    private val historyRepo: SearchHistoryRepository
+) {
+    suspend fun execute(query: String, filters: MediaFilters?, language: Language): List<TvSeries> {
+        historyRepo.addSearchHistoryItem(
+            SearchHistoryInputItem(
+                query = query,
+                category = SearchCategory.MOVIE,
```
ZainabAlnajjar: why SearchCategory is movie in the TvSeries search usecase, please change it

domain/vod/src/main/java/usecase/SearchActorsUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class SearchActorsUseCase(
+    private val searchRepository: SearchRepository,
+    private val historyRepo: SearchHistoryRepository
+) {
+    suspend fun execute(query: String, language: Language): List<Actor> {
+        historyRepo.addSearchHistoryItem(
+            SearchHistoryInputItem(
+                query = query,
+                category = SearchCategory.MOVIE,
```
ZainabAlnajjar: same here change to actor

domain/vod/src/main/java/repository/SearchRepository.kt
```diff
@@ -0,0 +1,21 @@
...
+import usecase.search.SearchActorHistoryOutput
+import usecase.search.SearchMediaHistoryOutput
+
+interface SearchRepository {
+    suspend fun searchActors(query: String, language: Language): List<SearchActorHistoryOutput>
+    suspend fun searchMovies(
+        query: String,
+        filters: MediaFilters?,
+        language: Language
+    ): List<SearchMediaHistoryOutput>
```
haideraljawasim: change the name to SearchActorOutput

ElhamFadel (dismissed):

domain/vod/src/main/java/entity/Genre.kt
```diff
@@ -0,0 +1,30 @@
...
+    ACTION,//
+    ADVENTURE,//
+    COMEDY,//
+    DRAMA,//
```
AimanYosofiUlt: What is the tow slash for?

domain/vod/src/main/java/entity/Genre.kt
```diff
@@ -1,30 +1,29 @@
...
+    HISTORY,
+    KIDS,
+    MYSTERY,
+    MUSIC,
+    NEWS,
+    REALITY,
+    SOAP,
+    TALK,
+    WAR,//   WESTERN,
+    TV_MOVIE,
```
haideraljawasim: I think you commented WESTERN by mistake

ZainabAlnajjar: my bad, fixed that

haideraljawasim (approved):

domain/vod/src/main/java/extensions/DateTimeExtensions.kt
```diff
@@ -0,0 +1,10 @@
+package extensions
+
+import kotlinx.datetime.Clock
+import kotlinx.datetime.LocalDateTime
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.toLocalDateTime
+
+fun LocalDateTime.Companion.now(
```
Ahmed7osny1, line 8: - i suggest to change now to current to avoid confusion with existing LocalDateTime.now()

domain/vod/src/main/java/repository/SearchRepository.kt
```diff
@@ -0,0 +1,21 @@
...
+        query: String,
+        filters: MediaFilters?,
+        language: Language
+    ): List<SearchMediaOutput>
+
+    suspend fun searchTvSeries(
+        query: String,
+        filters: MediaFilters?,
+        language: Language
+    ): List<SearchMediaOutput>
```
Ahmed7osny1, line 20: i think we have two type (tv and movies) and these two fun the same 
we can provide enum
enum class MediaType {
    MOVIE, TV_SERIES
}
and write one methods but take type from  MediaType

domain/vod/src/main/java/usecase/SearchActorsUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlinx.datetime.LocalDateTime
+import repository.SearchHistoryRepository
+import repository.SearchRepository
+import usecase.search.SearchActorOutput
+import usecase.search.SearchCategory
+import usecase.search.SearchHistoryInputItem
+
+class SearchActorsUseCase(
+    private val searchRepository: SearchRepository,
+    private val historyRepo: SearchHistoryRepository
```
Ahmed7osny1, line 14: i prefer to change this to historyRepository for consistancy

domain/vod/src/main/java/usecase/SearchMoviesUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+import repository.SearchHistoryRepository
+import repository.SearchRepository
+import usecase.search.MediaFilters
+import usecase.search.SearchCategory
+import usecase.search.SearchHistoryInputItem
+import usecase.search.SearchMediaOutput
+
+class SearchMoviesUseCase(
+    private val searchRepository: SearchRepository,
+    private val historyRepo: SearchHistoryRepository
```
Ahmed7osny1, line 16: i prefer to change this to historyRepository for consistancy

domain/vod/src/main/java/usecase/SearchTvSeriesUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+import repository.SearchHistoryRepository
+import repository.SearchRepository
+import usecase.search.MediaFilters
+import usecase.search.SearchCategory
+import usecase.search.SearchHistoryInputItem
+import usecase.search.SearchMediaOutput
+
+class SearchTvSeriesUseCase(
+    private val searchRepository: SearchRepository,
+    private val historyRepo: SearchHistoryRepository
```
Ahmed7osny1, line 16: i prefer to change this to historyRepository for consistancy

AimanYosofiUlt (approved):

domain/vod/src/main/java/entity/Language.kt
```diff
@@ -0,0 +1,6 @@
+package entity
+
+enum class Language(val code: String) {
```
Ahmed7osny1, line 3: we can use language provider

domain/vod/src/main/java/usecase/search/SearchHistory.kt
```diff
@@ -0,0 +1,9 @@
+package usecase.search
+
+import kotlinx.datetime.LocalDateTime
+
+data class SearchHistory (
+    val id: Long? = null,
```
Ahmed7osny1, line 6: why nullable

domain/vod/src/main/java/usecase/search/SearchHistoryInputItem.kt
```diff
@@ -0,0 +1,13 @@
+package usecase.search
+
+import entity.Language
+import kotlinx.datetime.LocalDateTime
+
+data class SearchHistoryInputItem(
+    val id: Long? = null,
```
Ahmed7osny1, line 7: same thing

domain/vod/src/main/java/usecase/search/SearchMediaOutput.kt
```diff
@@ -0,0 +1,8 @@
+package usecase.search
+
+data class SearchMediaOutput(
+    val id: Long? = null,
```
Ahmed7osny1, line 4: same thing

PR #37, opened 2025-07-09
implement searchResult caching

### LINKED ISSUE #26: [Search Feature] [Data] Create Room caching logic for Search
This component responsible to save the last result search .

ElhamFadel:


gradle/libs.versions.toml
```diff
@@ -82,6 +87,18 @@ koin-androidx-compose = { group = "io.insert-koin", name = "koin-androidx-compos
...
+# Room dependencies
+androidx-room-runtime = { group = "androidx.room", name = "room-runtime", version.ref = "room" }
+androidx-room-ktx = { group = "androidx.room", name = "room-ktx", version.ref = "room" }
+
+# Coroutines
+kotlinx-coroutines-android = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-android", version.ref = "coroutines" }
+
+# Hilt dependencies
+hilt-android = { group = "com.google.dagger", name = "hilt-android", version.ref = "hilt" }
+hilt-compiler = { group = "com.google.dagger", name = "hilt-compiler", version.ref = "hilt" }
```
haideraljawasim: why you added hilt? we gonna  koin for di  as required from us

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchResultDao.kt
```diff
@@ -0,0 +1,20 @@
...
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import com.sanaa.search.data.local.entity.SearchResultEntity
+
+@Dao
+interface SearchResultDao {
+
+    @Query("SELECT * FROM search_results WHERE `query` = :query AND `language` = :language")
+    suspend fun getByQueryAndLanguage(query: String, language: String): List<SearchResultEntity>
```
haideraljawasim: this logic not correct we also need to get data by other filters not just by query like release year or genre  using this in repository if the user wrote the same query with different filters will always get the same result like just from local data source , we need other filters to be matched like if there is result for the exact search input from query to filter we will return to from local data source otherwise  we have to call the api again

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/datasource/LocalSearchDataSourceImpl.kt
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    override suspend fun getCachedResults(query: String, language: String): List<SearchResultEntity> {
+        return dao.getByQueryAndLanguage(query, language)
+    }
+
+    override suspend fun clearExpiredResults() {
+        val oneHourAgo = System.currentTimeMillis() - 60 * 60 * 1000
+        dao.deleteOldResults(oneHourAgo)
+    }
```
haideraljawasim: clearing the cached data should  be automated from room data base we don't want to call a function to delete 
like if went with this approach where should this be called ?

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/entity/SearchResultEntity.kt
```diff
@@ -0,0 +1,15 @@
...
+@Entity(tableName = "search_results")
+data class SearchResultEntity(
+    @PrimaryKey val id: Int,
+    val query: String,
+    val language: String,
+    val timestamp: Long,
+    val mediaType: String, // movie / tv / actor
+    val titleOrName: String,
+    val imagePath: String?
+)
```
haideraljawasim, line 15: I see here a big problem you use the key of the movie or series as primary and i noticed you replace it on conflict in dao file , now suppose the user searched for item let say the query was "fight" the result here contained Fight Club movie but  later the user search "fight club" also the result contained Fight club movie here you replaced the old movie of "fight" query with this new,  again if the user search for just the word "fight" here will not get fight club movie since it have been replaced with new one with different query and you get the item by query so this not applicable   ,

also , storing each item from the search result along with its query will lead to a lot of redundant data.

I prefer going with table for the search and search filters with the ids of search result and another three table like one for each category to store the actual data like id and image and title add to language

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/db/AppDatabase.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.sanaa.search.data.local.dao.SearchResultDao
+import com.sanaa.search.data.local.entity.SearchResultEntity
+
+@Database(
+    entities = [SearchResultEntity::class],
+    version = 1,
+    exportSchema = false
+)
+abstract class AppDatabase : RoomDatabase() {
+    abstract fun searchResultDao(): SearchResultDao
```
haideraljawasim: Im not sure but i suppose we should have to initialize one Room Data Base for all the project which it should in App module or add Core data source module , we may ask the mentors about this later

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/entity/SearchResultEntity.kt
haideraljawasim: I see here a big problem you use the key of the movie or series as primary and i noticed you replace it on conflict in dao file , now suppose the user searched for item let say the query was "fight" the result here contained Fight Club movie but later the user search "fight club" also the result contained Fight club movie here you replaced the old movie of "fight" query with this new, again if the user search for just the word "fight" again , here will not get fight club movie since it have been replaced with new one with different query and you get the item by query so this not applicable ,

also , storing each item from the search result along with its query will lead to a lot of redundant data.

I prefer going with table for the search and search filters with the ids of search result and another three table like one for each category to store the actual data like id and image and title add to language

haideraljawasim: One more thing: please use annotations so that Room database recognizes them. for example, if a property has been renamed, add an annotation like this for each property:
` @ColumnInfo(name = "language")
    val language: String,`

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/datasource/LocalSearchDataSource.kt
```diff
@@ -0,0 +1,19 @@
...
+        results: List<SearchResultEntity>
+    )
+
+    suspend fun getCachedResults(
+        query: String,
+        language: String
+    ): List<SearchResultEntity>
+
+    suspend fun clearExpiredResults()
+}
```
haideraljawasim: I don’t see why we need an interface for our data source. Do we need to implement it in a different way? Please correct me if I’m wrong.

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/datasource/LocalSearchDataSourceImpl.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.sanaa.search.data.local.dao.SearchResultDao
+import com.sanaa.search.data.local.entity.SearchResultEntity
+
+class LocalSearchDataSourceImpl(
+    private val dao: SearchResultDao
+) : LocalSearchDataSource {
+
+    override suspend fun saveSearchResults(query: String, language: String, results: List<SearchResultEntity>) {
+        dao.insertAll(results)
+    }
```
haideraljawasim: here you pass the query and language but don't use it !!!

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/entity/SearchResultEntity.kt
```diff
@@ -0,0 +1,15 @@
+package com.sanaa.search.data.local.entity
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity(tableName = "search_results")
+data class SearchResultEntity(
```
haideraljawasim: In the public code review Barq recommended naming Room database entities as LocalDto to avoid conflicts with domain layer entities that are also named Entity

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchResultDao.kt
ElhamFadel: Yup, we have filter so we need another function handle coming parameters.

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/datasource/LocalSearchDataSourceImpl.kt
ElhamFadel: We need it but I forgot embedded it inside searchResult object 

Thanks Haider

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/datasource/LocalSearchDataSourceImpl.kt
ElhamFadel: hmm, I know 
But How we can did that automatically ?

gradle/libs.versions.toml
ElhamFadel: yes this my conflict

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/datasource/LocalSearchDataSource.kt
AhmadKharfan: I think the interface keeps us flexible Room today, CSV/DataStore/etc. tomorrow With the interface in place we can swap those implementations (or mock them) while the rest of the app stays unchanged

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/db/AppDatabase.kt
AhmadKharfan: Yes one shared Room DB with all DAOs
Elham already registers that DB as a singleton in Koin and create instance in the Koin so the whole project can inject the same instance

AhmadKharfan: If we add new entities/DAOs later just include them in this @Database declaration and add the abstract fun dao so they live in the same instance

haideraljawasim: Yes i know Elham did that but i was wondering cause the initialized of room data base now in search data source model

app/build.gradle.kts
```diff
@@ -103,4 +103,7 @@ dependencies {
     implementation(libs.ktor.serialization.kotlinx.json)
     implementation(libs.ktor.client.logging)
     implementation(libs.kotlinx.serialization.json)
+    
+    // Local Data Sources
+    implementation(project(":data:local_data_source:search"))
```
devfalah:
```suggestion
    implementation(projects.data.localDataSource.search)
```

shahad-97:
good job

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchDao.kt
```diff
@@ -0,0 +1,23 @@
...
+interface SearchDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertSearch(search: SearchLocalDto): Long
+
+    @Query("SELECT * FROM searches WHERE `query` = :query AND language = :language LIMIT 1")
+    suspend fun getSearchByQueryAndLanguage(query: String, language: String): SearchLocalDto?
+
+    @Query("SELECT * FROM searches ORDER BY timestamp DESC")
+    suspend fun getAllSearches(): List<SearchLocalDto>
```
Radwa85: return flow <list> for queries to enable reactive data updates

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchDao.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    @Query("SELECT * FROM searches WHERE `query` LIKE '%' || :query || '%' ORDER BY timestamp DESC")
+    suspend fun getSearchesByQuery(query: String): List<SearchLocalDto>
```
Radwa85: also here

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchDao.kt
```diff
@@ -0,0 +1,23 @@
...
+import androidx.room.Query
+import com.sanaa.search.data.local.entity.SearchLocalDto
+
+@Dao
+interface SearchDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertSearch(search: SearchLocalDto): Long
+
+    @Query("SELECT * FROM searches WHERE `query` = :query AND language = :language LIMIT 1")
```
Ahmed7osny1: you don't need to limit, and you don't need to language.
it will compare with query

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchResultDao.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertAll(results: List<SearchResultLocalDto>)
+
+    @Query("""
+        SELECT sr.* FROM search_results sr
+        INNER JOIN searches s ON sr.search_id = s.search_id
+        WHERE s.query = :query AND s.language = :language
+    """)
+    suspend fun getByQueryAndLanguage(query: String, language: String): List<SearchResultLocalDto>
```
Ahmed7osny1: use flow list to Reactive updates

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SeriesDao.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.sanaa.search.data.local.entity.SeriesLocalDto
+
+@Dao
+interface SeriesDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertSeries(series: SeriesLocalDto)
+
+    @Query("SELECT * FROM series WHERE release_year BETWEEN :minYear AND :maxYear AND genres LIKE :genre AND imdb_rating >= :rating")
+    suspend fun getFilteredSeries(minYear: Int, maxYear: Int, genre: String, rating: Float): List<SeriesLocalDto>
```
Ahmed7osny1: use flow list Reactive updates

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/datasource/LocalSearchDataSource.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.sanaa.search.data.local.entity.SearchResultLocalDto
+
+interface LocalSearchDataSource {
+
+    suspend fun saveSearchResult(
+        query: String,
+        language: String,
+        itemId: Int,
+        itemType: String
+    )
```
Ahmed7osny1: change name of function and parameter name

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/entity/ActorsLocalDto.kt
```diff
@@ -0,0 +1,18 @@
+package com.sanaa.search.data.local.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity(tableName = "actors")
```
Ahmed7osny1: change table name to actor, table name should not be as collect

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/entity/MoviesLocalDto.kt
```diff
@@ -0,0 +1,27 @@
+package com.sanaa.search.data.local.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity(tableName = "movies")
```
Ahmed7osny1: the same here

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/entity/SearchLocalDto.kt
```diff
@@ -0,0 +1,21 @@
+package com.sanaa.search.data.local.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity(tableName = "searches")
```
Ahmed7osny1: and here

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/entity/SearchResultLocalDto.kt
```diff
@@ -0,0 +1,28 @@
+package com.sanaa.search.data.local.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.ForeignKey
+
+@Entity(
+    tableName = "search_results",
```
Ahmed7osny1: and here

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/entity/SeriesLocalDto.kt
```diff
@@ -0,0 +1,27 @@
+package com.sanaa.search.data.local.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity(tableName = "series")
```
Ahmed7osny1: and here

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchDao.kt
ElhamFadel: I think we don't need it here

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchDao.kt
ElhamFadel: why?

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchResultDao.kt
ElhamFadel: we just call it once so why we need get updates ?

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchDao.kt
ZainabAlnajjar: we need the language because the result is different for each

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SeriesDao.kt
ZainabAlnajjar: we don't need it here

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/SearchResultDao.kt
ZainabAlnajjar: we don't need the flow here

data/local_data_source/search/src/main/java/com/sanaa/search/data/local/dao/ActorDao.kt
```diff
@@ -0,0 +1,17 @@
...
+import com.sanaa.search.dataSource.local.dto.ActorsLocalDto
+
+@Dao
+interface ActorDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertActor(actor: ActorsLocalDto)
+
+    @Query("SELECT * FROM actor WHERE name LIKE '%' || :query || '%' LIMIT 1")
+    suspend fun getActorByQuery(query: String): ActorsLocalDto?
```
haideraljawasim: since Room data base case sensitive
you should use this 
`@Query("SELECT * FROM actor WHERE LOWER(name) LIKE '%' || LOWER(:query) || '%' LIMIT 1")`

AhmadKharfan (dismissed):

shahad-97 (dismissed):

Radwa85 (dismissed):

shahad-97 (approved):

AhmadKharfan (approved):

PR #89, opened 2025-07-12
refactor search viewmodel

EL-MANCY:
refactor and enhance renaming for search viewmodel edit some files locations

### DIFF
--- a/feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchScreen.kt
+++ b/feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchScreen.kt
@@ -121,7 +121,7 @@ fun SearchScreenContent(
             )
             Spacer(Modifier.height(12.dp))
 
-            if (uiState.isLoading && uiState.searchQuery.isNotBlank()) {
+            if (uiState.isLoading) {
                 Box(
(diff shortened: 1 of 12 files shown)


haideraljawasim (approved):

feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchScreen.kt
```diff
@@ -82,60 +89,49 @@ fun SearchScreenContent(
...
+                    WavyProgressIndicator()
+                }
+            } else {
+                CategoryTabSection(
+                    selectedTabIndex = uiState.selectedTabIndex,
+                    onTabSelected = listener::onTabSelected,
+                    uiState = uiState,
+                    listener = listener
+                )
+            }
```
ZainabAlnajjar: why checking if the query is blank? I think we should only check for loading state

ZainabAlnajjar (approved):
great job Abdulrahman, the code is much cleaner and more readable now

feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchScreen.kt
EL-MANCY: discussed and resolved

PR #102, opened 2025-07-13
support arabic language

### LINKED ISSUE #94: localization to Arabic
localize everything left in the screen to Arabic language, including genres

ElhamFadel:


design_system/src/main/res/values/strings.xml
```diff
@@ -18,4 +18,38 @@
     <string name="apply">Apply</string>
     <string name="cancel">Cancel</string>
     <string name="search_hint">Search…</string>
+    
+    <!-- Tab Labels -->
+    <string name="movies">Movies</string>
```
devfalah, line 25: Why are you putting strings in the design system module?

design_system/src/main/res/values/strings.xml
```diff
@@ -18,4 +18,38 @@
...
+    <string name="genre_war">War</string>
+    <string name="genre_talk">Talk</string>
+    <string name="genre_soap">Soap</string>
+    <string name="genre_reality">Reality</string>
+    <string name="genre_news">News</string>
+    <string name="genre_tv_movie">TV Movie</string>
+    <string name="genre_western">Western</string>
+    <string name="genre_war_and_politics">War and Politics</string>
+    <string name="genre_sci_fi_and_fantasy">Sci-Fi and Fantasy</string>
+    <string name="genre_action_and_adventure">Action and Adventure</string>
```
ZainabAlnajjar, line 54: Maybe you could replace `and` with `&` to take less space and the user won't feel the chip is too long? just my opinion as a user would like it more

app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,37 @@
...
+    <string name="genre_war">War</string>
+    <string name="genre_talk">Talk</string>
+    <string name="genre_soap">Soap</string>
+    <string name="genre_reality">Reality</string>
+    <string name="genre_news">News</string>
+    <string name="genre_tv_movie">TV Movie</string>
+    <string name="genre_western">Western</string>
+    <string name="genre_war_and_politics">War and Politics</string>
+    <string name="genre_sci_fi_and_fantasy">Sci-Fi and Fantasy</string>
+    <string name="genre_action_and_adventure">Action and Adventure</string>
```
ZainabAlnajjar: Maybe you could replace `and` with `&` to take less space and the user won't feel the chip is too long? just my opinion as a user would like it more

ZainabAlnajjar (dismissed):
Nice work, just remove the genres strings from design system, we will pass the localized text when we call the chips from the app module

design_system/src/main/res/values/strings.xml
ElhamFadel: It's my conflict I was rush

design_system/src/main/res/values-ar/strings.xml
ZainabAlnajjar, line 1: I believe translation is not related to design system, but it's the app responsibility to provide it

It would be better to move all the translation to the app resources as what you did with the genres (also move these again to app resources please), so the translation would be the same throughout the app, and when we want to edit something, only that file would change (even when we want to support other languages)

raouf-kamel (approved):
🌟 **Great work, Elham**

AimanYosofiUlt (approved):

shahad-97 (approved):

haideraljawasim (approved):

Ahmed7osny1 (approved):

PR #104, opened 2025-07-13
Fix/unit test

### LINKED ISSUE #84: fix unit test
Fix some unit test that failed

AimanYosofiUlt:
Prepare unit test for the part that has not been tested before to increase the coverage from 9% to 95%


feature/search/presentation/src/main/java/com/sanaa/presentation/filter_bottomsheet/state/FilterUiState.kt
```diff
@@ -4,9 +4,9 @@ import entity.Genre
 
 data class FilterUiState(
     val yearRange: ClosedFloatingPointRange<Float> = 1980f..2025f,
-    val allGenres: List<Genre> = emptyList(),
+    val allGenres: List<Genre> = Genre.entries,
     val selectedGenres: Set<Genre> = emptySet(),
     val imdbRating: Int = 0,
-    val isLoading: Boolean = true,
+    val isLoading: Boolean = false,
```
Radwa85, line 10: i think We should keep it true by default and set it to false after loading finishes. Let me know if there's a specific reason for the change

feature/search/presentation/src/test/java/com/sanaa/presentation/filter_bottomsheet/FilterViewModelTest.kt
```diff
@@ -0,0 +1,129 @@
...
+            val expected = FilterUiState(
+                imdbRating = rate,
+                isDefaultState = false
+            )
+            Truth.assertThat(item).isEqualTo(expected)
+        }
+    }
+
+    @Test
+    fun `onClearFilters()`() = runTest {
```
Radwa85: using function name as name of test not enough , u can add more description

feature/search/presentation/src/test/java/com/sanaa/presentation/filter_bottomsheet/FilterViewModelTest.kt
```diff
@@ -0,0 +1,129 @@
...
+        // Then
+        filterViewModel.uiState.test {
+            val item = awaitItem()
+            val expected = FilterUiState(yearRange = range, isDefaultState = false)
+            Truth.assertThat(item).isEqualTo(expected)
+        }
+    }
+
+    @Test
+    fun `onGenreSelected() should change genre`() = runTest {
```
Radwa85: add when statement too

feature/search/presentation/src/main/java/com/sanaa/presentation/filter_bottomsheet/state/FilterUiState.kt
AimanYosofiUlt: There is no loading data happening in the initial

feature/search/presentation/src/test/java/com/sanaa/presentation/filter_bottomsheet/FilterViewModelTest.kt
```diff
@@ -0,0 +1,129 @@
...
+        // When
+        filterViewModel.onYearRangeChanged(range)
+
+        // Then
+        filterViewModel.uiState.test {
+            val item = awaitItem()
+            val expected = FilterUiState(yearRange = range, isDefaultState = false)
+            Truth.assertThat(item).isEqualTo(expected)
+        }
+    }
```
ZainabAlnajjar: I think it would be clearer to say it would update the state instead of changing the range, like this:
`onYearRangeChanged() should update filter state when called`, same for other tests

also, you cold import the `assertThat `instead of calling it from `Truth.assertThat` -> `assertThat`

ZainabAlnajjar (approved):

Radwa85 (approved):

mostafa-n3ma (approved):

shahad-97 (approved):

PR #105, opened 2025-07-13
block the bottom sheet from sliding while using the ReleaseYearSlider.kt

EL-MANCY:
fix some imports
block the bottom sheet from sliding while using the ReleaseYearSlider.kt

https://github.com/user-attachments/assets/12ff6573-8614-438f-b351-c7d96ba6554a


mostafa-n3ma (approved):
nice Jobe , thank you for your efforts

ZainabAlnajjar (approved):

PR #121, opened 2025-07-14
Feature/create details domain

### LINKED ISSUE #107: Create Domain Details

AhmadKharfan:
- Details domain added: ActorRepository, MovieRepository, TvSeriesRepository.

- Use-cases for every repo operation (get details, reviews, seasons, etc.).

- Unit tests for all use-cases → 100 % coverage.


domain/vod/src/main/java/details/repository/ActorRepository.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import entity.Actor
+import entity.Movie
+import entity.TvSeries
+
+interface ActorRepository {
+    suspend fun getActorDetails(id: Int): Actor
+    suspend fun getProfileImages(id: Int): List<String>
+    suspend fun getGalleryImages(id: Int): List<String>
+    suspend fun getTopMovies(id: Int): List<Movie>
```
mostafa-n3ma: getting the top movies or top series is the actor repo responsibility   !

mostafa-n3ma: if it the actors specific top movies maybe we should rename it to actorTopMovies or topRelativeMovies something indicates the actor involved 
but still we are getting movies data in ActorRepository 
I think this is weird

domain/vod/src/main/java/details/repository/MovieRepository.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import entity.Actor
+import entity.Genre
+import entity.Movie
+import entity.Review
+
+interface MovieRepository {
+    suspend fun getMovieDetails(id: Int): Movie
+    suspend fun getImages(id: Int): List<String>
+    suspend fun getCast(id: Int): List<Actor>
```
mostafa-n3ma: again I don't  fill right about getting a list of actors in a movieRepo ! 
maybe we can get topMovies with actor id from the movies repo and getting cast actors list by movie id  from actor repo

domain/vod/src/main/java/details/repository/ActorRepository.kt
Ahmed7osny1: I agree with mostafa, top movies related to actor we can use getActorTopMovies

domain/vod/src/main/java/details/repository/MovieRepository.kt
Ahmed7osny1: @mostafa-n3ma But I think cast related to movie, and the same thing top movies related to actor

domain/vod/src/main/java/details/usecase/movie/GetMoviesByCategory.kt
```diff
@@ -0,0 +1,10 @@
+package details.usecase.movie
+
+import details.repository.MovieRepository
+import entity.Genre
+import entity.Movie
+
+class GetMoviesByCategory(private val movieRepository: MovieRepository) {
+    suspend fun execute(category: Genre): List<Movie> =
+        movieRepository.getMoviesByCategory(category)
```
Ahmed7osny1, line 9: Take care about empty lines

Ahmed7osny1 (dismissed):
In general, greate work. 
But I refer to be consistency of naming in inject repo, I saw once repo, repository, actor repo.
We can discuss about approach use in all naming to achieve consistency.
Thanks for hard work and quick delivery, Keep going  ♥️💪

domain/vod/src/main/java/details/repository/MovieRepository.kt
mostafa-n3ma: I knew that repo can have multiple data sources and manages them but it should be for the same data type right .
like remot and local datasources for actor type 
Not data sources for different types 
Like using actor local and remote data sources and movies local and remote data sources all in one repo , imagine testing that repo  !!

shahad-97 (dismissed):
well done

domain/vod/src/main/java/details/repository/ActorRepository.kt
AhmadKharfan: Actually it's inside the ActorRepository when we call it as actorRepository.getTopMovies(), it feels related to the actor context But your suggestion also makes sense for better readability

domain/vod/src/main/java/details/repository/MovieRepository.kt
AhmadKharfan: I think the cast is related to the movie, so I don't see a problem putting it inside the MovieRepository
The return type doesn’t have to match the repository name exactly the repo's responsibility is tied to the data's source or context not just the object it returns

AhmadKharfan (commented):
@Ahmed7osny1 @mostafa-n3ma 
Thanks for your valuable comments to improve readability
I’ll work on it to ensure more consistency

shahad-97 (approved):

Ahmed7osny1 (approved):

PR #138, opened 2025-07-14
Implement Decoupled Multi-Module Navigation

EL-MANCY:
### Key Architectural Changes

We've completely reimagined how navigation works in the app. Think of it less like a tangled web of roads and more like a modern airport with a central control tower. Here’s how the new system is organized:

---

### The "How to Get There" Manuals  (`:api` modules)

Each feature (like Home, Search, etc.) now comes with its own small, public "manual" or `:api` module.

* **What it does:** This manual tells every other part of the app the feature's public address (its navigation `route`).
* **No secrets:** It contains no UI or secret logic. It just provides the public information needed to find it.
* **Safe shortcuts:** If a feature needs extra info to open (like a movie ID), the manual provides a safe helper function (e.g., `buildRoute(mediaId)`) to make sure nobody gets the address wrong.

---

### The Self-Contained Islands (`:presentation` modules)

Each feature's screen is now like its own isolated island.

* **What it does:** The island holds all the UI and logic for that single feature.
* **Knows nothing about the outside world:** The Search island has no idea how to get to the Home island. It's completely self-contained.
* **Sends up flares:** When a user wants to go somewhere else, the screen doesn't try to navigate itself. Instead, it sends up a "flare" (a callback) that says something specific like, *"Hey, the user just clicked on a movie!"* or *"The user wants to go back!"*

---

### The Air Traffic Controller (`:app` module)

The `:app` module acts as the central air traffic controller for the entire application.

* **What it does:** It has a new `AppNavigation` component that watches for all the "flares" sent up from the feature islands.
* **Connects the dots:** When it sees a flare, it knows exactly what to do. If the Home island sends a *"user clicked a movie"* flare, the controller gives it the correct flight path to the Media Details island.
* **The single source of truth:** All navigation logic lives here. This makes it incredibly easy to see how all the features are connected and to manage the flow of the entire app from one place.

### DIFF
--- a/app/src/main/java/com/sanaa/navigation/AppNavigation.kt
+++ b/app/src/main/java/com/sanaa/navigation/AppNavigation.kt
@@ -1,0 +1,106 @@
+package com.sanaa.navigation
+
+import androidx.compose.runtime.Composable
+import androidx.navigation.NavHostController
+import androidx.navigation.compose.NavHost
+import androidx.navigation.compose.rememberNavController
+import com.sanaa.api.AuthenticationFeatureApi
+import com.sanaa.api.HomeFeatureApi
+import com.sanaa.api.MediaDetailsFeatureApi
+import com.sanaa.api.OnboardingFeatureApi
+import com.sanaa.api.PlaylistsFeatureApi
+import com.sanaa.api.SavedContentFeatureApi
+import com.sanaa.api.SearchFeatureApi
+import com.sanaa.api.UserProfileFeatureApi
+import com.sanaa.presentation.navigation.registerAuthenticationScreen
+import com.sanaa.presentation.navigation.registerHomeScreen
+import com.sanaa.presentation.navigation.registerMediaDetailsScreen
+import com.sanaa.presentation.navigation.registerOnboardingScreen
+import com.sanaa.presentation.navigation.registerPlaylistsScreen
+import com.sanaa.presentation.navigation.registerSavedContentScreen
+import com.sanaa.presentation.navigation.registerSearchScreen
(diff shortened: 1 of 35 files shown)


feature/authentication/api/src/main/java/com/sanaa/api/AuthenticationFeatureApi.kt
```diff
@@ -0,0 +1,5 @@
+package com.sanaa.api
+
+object AuthenticationFeatureApi {
+    const val route: String = "authentication_route"
+}
```
ZainabAlnajjar, line 5: please use enum class for the routes, that would be safer

app/src/main/java/com/sanaa/navigation/AppNavigation.kt
```diff
@@ -0,0 +1,106 @@
...
+import com.sanaa.presentation.navigation.registerSearchScreen
+import com.sanaa.presentation.navigation.registerUserProfileScreen
+
+
+@Composable
+fun AppNavigation(
+    navController: NavHostController = rememberNavController(),
+    startDestination: String = HomeFeatureApi.route
+) {
+    NavHost(
```
iBareq, line 30: with our structure we should enforce library to handle the navigation in all feature 
every feature could use different navigation library and the app module shouldn't contain any navigation code

iBareq (changes requested):
Rejected please re implement it based on the previous comment

app/src/main/java/com/sanaa/navigation/AppNavigation.kt
EL-MANCY: okay

PR #145, opened 2025-07-15
add TvSeriesRepositoryImplTest

shahad-97:


domain/vod/src/main/java/entity/Actor.kt
```diff
@@ -12,8 +12,7 @@ data class Actor(
     val character: String?,
     val birthDate: LocalDate?,
     val deathDate: LocalDate?,
-    val placeOfBirth: String?,
-    val biography: String
```
haideraljawasim, line 16: I don't get it why you removed  the biography from Actor ?

gradle/libs.versions.toml
```diff
@@ -146,6 +147,7 @@ kotlinx-datetime = { module = "org.jetbrains.kotlinx:kotlinx-datetime", version.
 timber = { module = "com.jakewharton.timber:timber", version.ref = "timber" }
 tensorflow-lite-task-vision = { module = "org.tensorflow:tensorflow-lite-task-vision", version.ref = "tensorflowLiteTaskVision" }
 turbine = { module = "app.cash.turbine:turbine", version.ref = "turbine" }
+places = { group = "com.google.android.libraries.places", name = "places", version.ref = "places" }
```
haideraljawasim, line 150: you added this but didn't use it

PR #150, opened 2025-07-15
add the core navigation between search and media details modules

EL-MANCY:


feature/media_details/api/src/main/java/com/sanaa/api/MediaDetailsApi.kt
```diff
@@ -0,0 +1,7 @@
+package com.sanaa.api
+
+import android.content.Context
+
+interface MediaDetailsApi {
+    fun launch(context: Context, mediaId: Int)
```
mostafa-n3ma, line 6: I think the context should be injected in the implementation not here in the api

ZainabAlnajjar:
the needed changes were done in another PR, we don't need this PR anymore

PR #158, opened 2025-07-16
Enable ProGuard & ABI Filtering to Reduce APK Size (~24.1 MB) 

### LINKED ISSUE #126: Reduce APK Size for Release Builds
Our current signed APK size (~50MB) is too large for efficient distribution via Firebase. To improve performance, user experience, and upload/download time, we need to analyze and reduce the release APK size.

Since we’re limited to APK (not AAB) due to using Firebase Distribution on the free tier, we’ll optimize the build using several strategies within the APK system.

Goals:
	•	Decrease the release APK size as much as possible
	•	Avoid packaging unused code, resources, or architectures
	•	Maintain or improve performance and stability

Ideas and listed Tasks:
	1.	✅ Enable R8/ProGuard and configure it properly:
	•	Set minifyEnabled true
	•	Add necessary proguard-rules.pro
	•	Prevent breaking Firebase or other SDKs with keep rules
	2.	✅ Enable resource shrinking

  3.✅ Convert all large .png and .jpg images to WebP format where applicable
	4.	✅ Run Analyze APK to inspect large files (fonts, drawables, .so, etc.)
	5.	✅ Review and remove unused resources (images, layouts, strings)
	6.	✅ Review and remove unused dependencies from build.gradle files
	7.	✅ Optimize module inclusion:
	•	Prevent incomplete or unnecessary modules from being packaged in the APK
	•	Use Gradle conditions or refactor modules to be optional in release builds
	8.	✅ Enable ABI splits in build.gradle:
	•	Only include "armeabi-v7a" and "arm64-v8a"
	•	Disable x86 and x86_64 unless needed

mostafa-n3ma:
✅ Changes Implemented:
•	Enabled ProGuard in the release build type of the app module to perform code shrinking, obfuscation, and optimization.
•	Fixed missing classes and applied appropriate -dontwarn rules to resolve R8-related issues.
•	Added ABI filters to include only:
•	arm64-v8a (64-bit ARM)
•	armeabi-v7a (32-bit ARM)
This eliminates unused native libraries and significantly reduces APK size.
•	Resulting APK size is now approximately 24.1 MB (was larger before these optimizations).

⸻

📌 Notes:
•	This change does not impact debug builds or functionality.
•	x86/x86_64 ABIs are intentionally excluded to minimize size—those are rarely used in production devices.
•	Preview-only resources were evaluated; unused ones will be addressed in a future cleanup task if needed.

⸻

Let me know if you’d like to include the APK size before the change for comparison!


gradle/libs.versions.toml
```diff
@@ -166,6 +166,9 @@ kover = { id = "org.jetbrains.kotlinx.kover", version.ref = "kover" }
 room = [
     "androidx-room-runtime",
     "androidx-room-ktx",
+]
+
+room-testing = [
```
mosayed01, line 171: No need to be a bundle anymore if it's only one dependency

mosayed01:
We should explore a way to apply shared android configurations (like ProGuard, shrinkResources, and buildFeatures) across all modules that use the Android plugin.
Right now, we’re repeating the same setup in every build.gradle, which isn’t scalable.

@mostafa-n3ma – Can you look into how we can centralize this setup so we don't have to manually configure each Android module?

devfalah:
> We should explore a way to apply shared android configurations (like ProGuard, shrinkResources, and buildFeatures) across all modules that use the Android plugin. Right now, we’re repeating the same setup in every build.gradle, which isn’t scalable.
> 
> @mostafa-n3ma – Can you look into how we can centralize this setup so we don't have to manually configure each Android module?

@mosayed01  Hosney and Zainab are working on this part and will finalize it soon

Ahmed7osny1:
> We should explore a way to apply shared android configurations (like ProGuard, shrinkResources, and buildFeatures) across all modules that use the Android plugin. Right now, we’re repeating the same setup in every build.gradle, which isn’t scalable.
> 
> @mostafa-n3ma – Can you look into how we can centralize this setup so we don't have to manually configure each Android module?

@mosayed01 I work on this

Ahmed7osny1 (approved):

ZainabAlnajjar (approved):

PR #177, opened 2025-07-17
Fix/search filters not working properly

### LINKED ISSUE #174: Fix search filters not working properly
- Filters not working properly (missing a lot of media in filter result)
- When applying filters, search won't refresh until you switch tab then return

ZainabAlnajjar:
- [x] Filters not working properly (missing a lot of media in filter result)
- [x] When applying filters, search won't refresh until you switch tab then return


data/repositories/search/src/main/java/com/sanaa/search/repository/SearchRepositoryImpl.kt
```diff
@@ -178,9 +165,9 @@ class SearchRepositoryImpl(
     private suspend fun <T> searchOrThrow(query: String, callee: suspend () -> T): T {
```
hagarabobakr: I think if the function name is 'safeCall' it will be more readable.

data/repositories/search/src/main/java/com/sanaa/search/repository/SearchRepositoryImpl.kt
```diff
@@ -35,14 +34,9 @@ class SearchRepositoryImpl(
...
                 }
-                return actors.map { it.toSearchOutput() }
+                actors.map { it.toSearchOutput() }
             }
-        } catch (_: UnknownHostException) {
-            throw NoNetworkException()
-        } catch (_: Exception) {
-            throw RetrievingDataFailureException("Failed to retrieve actors for query: $query")
         }
-    }
```
hagarabobakr: Suggestion: Move the filtering logic from searchMovies and searchTvShows to shared private functions within the file.

ZainabAlnajjar: @AimanYosofiUlt did that in his branch, my branch is to fix the logic only, we will combine them

feature/search/presentation/src/main/java/com/sanaa/presentation/screen/componants/SearchHistoryContent.kt
```diff
@@ -114,9 +114,7 @@ fun SearchHistoryContent(
                                         }
                                     },
                                     topLeftContent = {
-                                        SaveIconChip(
-                                            onClick = interactionsListener::onSaveIconClicked
-                                        )
+                                        SaveIconChip(onClick = {})
```
hagarabobakr: Why did we delete 
onClick = interactionsListener::onSaveIconClicked
 from here?

hagarabobakr (dismissed):
Good job 👏

data/repositories/search/src/main/java/com/sanaa/search/repository/SearchRepositoryImpl.kt
```diff
@@ -52,24 +46,22 @@ class SearchRepositoryImpl(
...
-                }
+                filteredMedia =
+                    filteredMedia.filter {
+                        (it.imdbRating ?: 0f) >= filters.imdbRating
+                                && it.releaseYear != null
+                                && it.releaseYear >= filters.startYear
+                                && it.releaseYear <= filters.endYear
+                    }
 
                 filteredMedia.map { it.toSearchOutput() }
```
hagarabobakr: This result is not returned from the function  you need to return it using return.

data/repositories/search/src/main/java/com/sanaa/search/repository/SearchRepositoryImpl.kt
```diff
@@ -88,20 +80,18 @@ class SearchRepositoryImpl(
...
+                    movies.filter { series ->
+                        val genreIds = series.genreIds?.toSet() ?: emptySet()
+                        filterGenreIds.all { it in genreIds }
+                    }
 
-                filters.imdbRating?.let { rating ->
-                    filteredMedia = filteredMedia.filter { (it.voteAverage ?: 0f) >= rating }
+                filteredMedia = filteredMedia.filter {
+                    (it.voteAverage ?: 0f) >= filters.imdbRating
+                            && it.releaseDate != null &&
```
hagarabobakr: LocalDate.parse is being called twice on the same value within the same condition. It's better to parse it once and store the result in a variable like:
val year = it.releaseDate?.let { date -> LocalDate.parse(date).year }
This improves performance slightly and enhances readability ^^

shahad-97 (dismissed):

data/repositories/search/src/main/java/com/sanaa/search/repository/SearchRepositoryImpl.kt
ZainabAlnajjar: it is returning the last line in each condition

feature/search/presentation/src/main/java/com/sanaa/presentation/screen/componants/SearchHistoryContent.kt
ZainabAlnajjar: it is failing the tests without any useful usage, we can implement it later

shahad-97 (approved):

AimanYosofiUlt (approved):

PR #193, opened 2025-07-18
fix crashing : cannot inject searchViewModel due to un available language provider instance

mostafa-n3ma:


raouf-kamel (approved):
good

ZainabAlnajjar (approved):

PR #210, opened 2025-07-19
Refactor/ create helper function to parse image urls

### LINKED ISSUE #209: Create helper function to parse image URLs ⏱️ 00:49:00

AhmadKharfan:
Use mapper helper to build full image URLs instead of manual string concatenation centralizes base URL logic and reduces duplication.


raouf-kamel (dismissed):
Please check the failing CI checks

data/repositories/series/src/test/java/com/sanaa/series/MapperTest.kt
```diff
@@ -202,7 +202,7 @@ class MapperTest {
 
     @Test
     fun `getProfileImageUrl formats URL correctly`() {
-        val url = getProfileImageUrl("/path.jpg")
+        val url = getFullImageUrl("/path.jpg")
```
ZainabAlnajjar: rename the test function also (line 204)

ZainabAlnajjar: also you have changed the base line to 
`private const val TMDB_IMAGE_BASE_URL = "https://image.tmdb.org/t/p/w500"`
I think this test is failing

data/repositories/search/src/main/java/com/sanaa/search/mapper/SearchMediaMapper.kt
```diff
@@ -60,8 +60,6 @@ fun TvShowSearchDto.toSearchOutput(): SearchTvSeriesOutput {
     return SearchTvSeriesOutput(
         id = id,
         title = name ?: "",
-        posterImageUrl = (IMAGE_URL + posterImagePath),
+        posterImageUrl = getFullImageUrl(posterImagePath),
     )
-}
-
-const val IMAGE_URL = "https://image.tmdb.org/t/p/original"
```
ZainabAlnajjar, line 67: the base url here is different from the one in the function you used, please double check the image you are getting and where it is displayed

data/repositories/series/src/test/java/com/sanaa/series/MapperTest.kt
ZainabAlnajjar: also please add the other usecase if the path is empty

ZainabAlnajjar (changes requested):
please check the base url you have chosen and if it gives the right image to display in each part of the app (media card, episodes, actors profile or gallery)

also please after that check the tests again, they shouldn't be passing, and rename the function in the test title

data/repositories/search/src/main/java/com/sanaa/search/mapper/SearchMediaMapper.kt
AhmadKharfan: I already tested the base URL and the only difference is in the image scale In other features we’re using the same link structure with w500 so I’d like to keep all base links that fetch movie or series posters using the same resolution or scale

AimanYosofiUlt (dismissed):

hagarabobakr (dismissed):

mostafa-n3ma (dismissed):

ZainabAlnajjar (dismissed):

AimanYosofiUlt (approved):

mostafa-n3ma (approved):

PR #214, opened 2025-07-20
Refactor/compain use case classes

### LINKED ISSUE #205: compain use case classes

AimanYosofiUlt:


app/src/main/java/com/sanaa/novix/di/domainModule.kt
```diff
@@ -5,28 +5,15 @@ import details.usecase.ManageEpisodeDetailsUseCase
...
-    single { SearchTvSeriesUseCase(get(), get()) }
-    single { SearchActorsUseCase(get(), get()) }
-    single { GetRecentViewedUseCase(get()) }
-    single { GetSearchHistoryUseCase(get()) }
-    single { ClearRecentViewedUseCase(get()) }
-    single { ClearSearchHistoryUseCase(get()) }
-    single { RemoveSearchHistoryUseCase(get()) }
-    single { AddRecentViewedUseCase(get()) }
-    single { RemoveSearchHistoryUseCase(get()) }
+    single { SearchUseCase(get(), get()) }
```
AhmadKharfan: You can use singleOf I think that would be better

domain/vod/src/main/java/search/usecase/SearchUseCase.kt
```diff
@@ -0,0 +1,32 @@
+package search.usecase
+
+import search.repository.SearchHistoryRepository
+import search.repository.SearchRepository
+import search.usecase.search_param.MediaFilters
+import search.usecase.search_param.SearchActorOutput
+import search.usecase.search_param.SearchMovieOutput
```
AhmadKharfan, line 7: We’re combining related use cases into one class so there’s no need for a separate `search_param` package please remove it because The data classes can be defined as inner classes if needed

domain/vod/src/main/java/search/usecase/SearchUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+import search.usecase.search_param.SearchTvSeriesOutput
+
+class SearchUseCase(
+    private val searchRepository: SearchRepository,
+    private val searchHistoryRepository: SearchHistoryRepository,
+) {
+    suspend fun searchMovies(query: String, page: Int, filters: MediaFilters?): List<SearchMovieOutput> {
```
AhmadKharfan, line 14: No need for separate `SearchMovieOutput`, `SearchActorOutput`, `SearchTvSeriesOutput` We can return the full data model (Actor, Movie, TvSeries) and map them in the UI layer to whatever we need

AhmadKharfan: So please update the repository and usecases to return full data class model no need to ouptut dataclass

domain/vod/src/main/java/search/usecase/ManageRecentViewedUseCase.kt
```diff
@@ -0,0 +1,23 @@
...
+    suspend fun addRecentViewed(item: RecentViewedMedia) = historyRepo.addRecentViewedMedia(item)
+    suspend fun clearRecentViewed() = historyRepo.clearRecentViewed()
+    suspend fun getRecentViewed(): Flow<List<RecentViewedMedia>> = historyRepo.getRecentViewed(RECENT_MEDIA_SIZE_LIMIT)
+
+    data class RecentViewedMedia(
+        val id: Int,
+        val posterImageUrl: String,
+        val mediaType: MediaType,
+        val isSaved: Boolean,
+    )
```
ElhamFadel, line 19: Can you split this outside

domain/vod/src/main/java/search/usecase/ManageRecentViewedUseCase.kt
```diff
@@ -0,0 +1,23 @@
...
+import search.repository.SearchHistoryRepository
+import search.usecase.search_param.MediaType
+
+class ManageRecentViewedUseCase(
+    private val historyRepo: SearchHistoryRepository,
+) {
+    suspend fun addRecentViewed(item: RecentViewedMedia) = historyRepo.addRecentViewedMedia(item)
+    suspend fun clearRecentViewed() = historyRepo.clearRecentViewed()
+    suspend fun getRecentViewed(): Flow<List<RecentViewedMedia>> = historyRepo.getRecentViewed(RECENT_MEDIA_SIZE_LIMIT)
+
```
ElhamFadel, line 13: ```RECENT_MEDIA_SIZE_LIMIT``` pass it as a parameter with a default value.

ElhamFadel (changes requested):

mostafa-n3ma:
ahmed and ilham , wrote a very valuable reviews and I agree with them all ,
please consider checking there comments

domain/vod/src/main/java/search/usecase/ManageRecentViewedUseCase.kt
ZainabAlnajjar: @ElhamFadel actually this item is only related to this useCase, so it is better to add it in the same file

domain/vod/src/main/java/search/usecase/SearchUseCase.kt
ZainabAlnajjar: we actually need to refactor the actor entity into 2 data classes, and replace the output model, I think it is better to leave it in separate branch dedicated to refactor entities

domain/vod/src/main/java/search/usecase/SearchUseCase.kt
AimanYosofiUlt: Sorry, but not in this PRs

domain/vod/src/main/java/search/usecase/SearchUseCase.kt
AimanYosofiUlt: Sorry, but not in this PRs

domain/vod/src/main/java/search/usecase/ManageRecentViewedUseCase.kt
AimanYosofiUlt: I don't think the viewModel should pass the limit

ElhamFadel (dismissed):

mostafa-n3ma (dismissed):

raouf-kamel (dismissed):

ZainabAlnajjar (approved):

haideraljawasim (approved):

PR #218, opened 2025-07-20
add items key in lazy column/row, so only the wanted items get re-com…

### LINKED ISSUE #217: always add items key in lazy column/row, so only the wanted items get re-composition, also to add animation on replacement
always add items key in lazy column/row, so only the wanted items get re-composition, also to add animation on replacement

AimanYosofiUlt:
…position, also to add animation on replacement

### DIFF
--- a/feature/media_details/presentation/src/main/java/com/sanaa/presentation/screen/actor/screen/ActorGalleryScreen.kt
+++ b/feature/media_details/presentation/src/main/java/com/sanaa/presentation/screen/actor/screen/ActorGalleryScreen.kt
@@ -41,7 +41,7 @@ import org.koin.core.parameter.parametersOf
 fun ActorGalleryScreen(
     actorId: Int,
     navigateBack: () -> Unit,
-    viewModel: ActorViewModel = koinViewModel { parametersOf(actorId) }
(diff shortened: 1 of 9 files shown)


ElhamFadel (approved):

feature/media_details/presentation/src/main/java/com/sanaa/presentation/screen/actor/screen/ActorGalleryScreen.kt
```diff
@@ -111,7 +111,9 @@ private fun ActorGalleryContent(
                                 8.dp
                             )
                         ) {
-                            items(state.galleryImageUrls) { image ->
+                            items(
+                                state.galleryImageUrls,
```
ZainabAlnajjar, line 115: I think you forgot to add key on this one

ZainabAlnajjar (approved):

feature/media_details/presentation/src/main/java/com/sanaa/presentation/screen/actor/screen/ActorGalleryScreen.kt
AimanYosofiUlt: I don't forget, it's not needed here, the value of it is the key

PR #219, opened 2025-07-20
applies common settings for Android application library modules

Ahmed7osny1:


build-logic/conventions/src/main/kotlin/Project.kt
ZainabAlnajjar, line 1: nice work, we also need to add test dependencies, as we are using them in all modules 

```
    testImplementation(libs.junit)
    androidTestImplementation(libs.androidx.junit)
    testRuntimeOnly(libs.junit.jupiter.engine)
    testImplementation(libs.mockk)
    testImplementation(libs.kotlinx.coroutines.test)
    testImplementation(libs.truth)
    testImplementation(kotlin("test"))
```

Ahmed7osny1: i think this now not in all modules.

ZainabAlnajjar: ok, I will check later and add them if we see things changed

data/remote_data_source/actors/build.gradle.kts
```diff
@@ -2,47 +2,27 @@ import java.io.FileInputStream
...
     }
 
     buildFeatures {
         buildConfig = true
     }
 }
 
+
 dependencies {
     implementation(projects.envConfig)
```
AimanYosofiUlt: This need to change: 
```suggestion
    implementation(projects.preferences)
```

devfalah (approved):

data/remote_data_source/actors/build.gradle.kts
ZainabAlnajjar: it will be changed after merging to div

raouf-kamel (approved):
🚀 **Impressive, Ahmed**

PR #224, opened 2025-07-20
Refactor/centralize koin di setup to app module

### LINKED ISSUE #223: Centralize Koin DI Setup to App Module

AhmadKharfan:
- Moved Koin initialization to the app module to centralize dependency injection
- Removed Koin modules from other modules 
- Ensured all DI modules are loaded from the app module


app/src/main/java/com/sanaa/novix/di/details_modules/remoteDetailsDataSource.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.sanaa.series.data_source.remote.RemoteTvSeriesDataSource
+import org.koin.dsl.module
+
+val remoteDetailsDataSource = module {
+    single<ActorRemoteDataSource> {
+        ActorRemoteDataSourceImpl(get(), get())
+    }
+    single<MovieDetailsRemoteDataSource> {
+        MovieDetailsRemoteDataSourceImpl(get(), get())
+    }
```
haideraljawasim: It would be better to update this to use singleOf with binding, since we’ve already started using it in some files.

app/src/main/java/com/sanaa/novix/di/search_modules/viewModelSearchModule.kt
```diff
@@ -0,0 +1,31 @@
...
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            dispatcher = Dispatchers.IO,
+        )
+    }
+    single { Dispatchers.IO }
```
haideraljawasim: same use `viewModelOf`

app/src/main/java/com/sanaa/novix/di/details_modules/remoteDetailsDataSource.kt
AhmadKharfan: I agree with you but in this PR I'm just moved the modules to the app module
So ok I will update it to use singleOf

app/src/main/java/com/sanaa/novix/di/search_modules/LocalDatabaseModule.kt
haideraljawasim, line 1: I noticed that this is the only file that starts with camel case, It would be better to rename it to align with the naming convention used in the other files. Also, other files include a suffix that explains in which feature used, but this one doesn’t, adding a suffix here would improve consistency

haideraljawasim (dismissed):

app/src/main/java/com/sanaa/novix/di/appModule.kt
```diff
@@ -1,18 +1,12 @@
...
+import com.sanaa.novix.di.details_modules.detailsModule
+import com.sanaa.novix.di.search_modules.searchModule
 import org.koin.dsl.module
 
 val appModule = module {
     includes(
-        firebaseModule, loggingModule, repositoryModule, localDatabaseModule,
-        networkModule, remoteDataSource, remoteDetailsDataSource, remoteMovieDetailsDataSource,
-        remoteTvSeriesDataSourceModule,preferencesModule, domainModule, navigationModule
+        firebaseModule, loggingModule, navigationModule, networkModule, preferencesModule,
```
mosayed01:
```suggestion
        firebaseModule, loggingModule, apiModule, networkModule, preferencesModule,
```

mosayed01: And inside apiModule use factory of instead of single of

mosayed01 (approved):

haideraljawasim (approved):

PR #232, opened 2025-07-21
fix: them of status bar

ElhamFadel:


app/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Dark Theme Colors -->
+    <color name="dark_background">#0D0608</color>
+    <color name="dark_surface">#0D0608</color>
+</resources>
```
ZainabAlnajjar: since you are using day and night theme and color, no need to change the name of the color, it is the same as localizing the string translation
just name it `background `and `surface `in both files with different values, and use them directly inside the theme

ZainabAlnajjar: also you are not using the `surface`

app/src/main/res/values/colors.xml
```diff
@@ -1,2 +1,10 @@
...
\ No newline at end of file
+<resources>
+    <!-- Dark Theme Colors -->
+    <color name="dark_background">#0D0608</color>
+    <color name="dark_surface">#0D0608</color>
+
+    <!-- Light Theme Colors -->
+    <color name="light_background">#FCF9F5</color>
+    <color name="light_surface">#FCF9F5</color>
+</resources>
```
ZainabAlnajjar: same applies here, only one color name will remain

ZainabAlnajjar:
also please remove the` SetNavigationBarColor` composable in `searchScreen`, it is not needed anymore

ZainabAlnajjar (approved):

haideraljawasim (approved):

PR #282, opened 2025-07-24
Task/implement web view

ElhamFadel:


feature/authentication/presentation/src/main/java/com/sanaa/presentation/navigation/AuthNavHost.kt
```diff
@@ -0,0 +1,55 @@
...
+import androidx.navigation.compose.rememberNavController
+import androidx.navigation.navArgument
+import com.sanaa.designsystem.design_system.theme.NovixTheme
+import com.sanaa.presentation.webview.ForgetPasswordWebViewRoute
+import com.sanaa.presentation.webview.SignUpWebViewRoute
+import com.sanaa.presentation.webview.SignUpWebViewScreen
+import com.sanaa.presentation.webview.ResetPasswordWebViewScreen
+
+@Composable
+fun AuthNavHost() {
```
EL-MANCY, line 18: What if we used a sealed class for the routes to avoid string typos, and created a helper function to cut down on the repeated code in the NavHost?

EL-MANCY (dismissed):
Great job elham, I suggested a refactor u can check it but anyways I will approve ur work

PR #291, opened 2025-07-25
Refactor/search domain

### LINKED ISSUE #289: Refactor search domain
Return movie and series entity instead of SearchOutput

haideraljawasim:
# What have been done :
- Delete output classes from search 
- Make search use case return entity 
- Delete default parameter from Media fillers

### DIFF
--- a/data/repositories/vod/src/main/java/com/sanaa/vod/mapper/search/SearchActorMapper.kt
+++ b/data/repositories/vod/src/main/java/com/sanaa/vod/mapper/search/SearchActorMapper.kt
@@ -13,7 +13,7 @@ fun ActorLocalDto.toEntity(): Actor {
         imageUrl = getFullImageUrl(imagePath),
         region = null,
         lastShow = null,
-        gender = Actor.Gender.MALE,
+        gender = mapGender(gender),
         department = null,
         character = null,
         birthDate = null,
@@ -29,6 +29,7 @@ fun ActorSearchDto.toLocalDto(language: String): ActorLocalDto {
         name = name ?: "",
         imagePath = getFullImageUrl(profileImagePath),
         language = language,
+        gender = gender
     )
 }
 
@@ -39,8 +40,8 @@ fun ActorSearchDto.toEntity(): Actor {
         imageUrl = getFullImageUrl(profileImagePath),
         region = null,
         lastShow = null,
-        gender = Actor.Gender.MALE,
-        department = null,
+        gender = mapGender(gender),
+        department = knownForDepartment,
         character = null,
         birthDate = null,
         deathDate = null,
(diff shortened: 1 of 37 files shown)


hagarabobakr (approved):
Thank you for making the search domainmore readable 👏

data/repositories/vod/src/main/java/com/sanaa/vod/mapper/search/SearchMediaMapper.kt
```diff
@@ -22,25 +28,36 @@ fun MovieSearchDto.toLocalDto(language: String): MovieLocalDto {
...
     )
 }
 
-fun MovieSearchDto.toSearchOutput(): SearchMovieOutput {
-    return SearchMovieOutput(
+fun MovieSearchDto.toEntity(): Movie {
+    return Movie(
         id = id,
-        title = title ?: "",
+        title = title.orEmpty(),
```
Radwa85, line 40: Is it expected that the API returns an empty title?

data/repositories/vod/src/main/java/com/sanaa/vod/mapper/search/SearchActorMapper.kt
```diff
@@ -20,16 +29,30 @@ fun ActorSearchDto.toLocalDto(language: String): ActorLocalDto {
         name = name ?: "",
```
Radwa85, line 29: and here too name of actor can be empty?

AimanYosofiUlt (approved):

data/repositories/vod/src/main/java/com/sanaa/vod/mapper/search/SearchMediaMapper.kt
haideraljawasim: Yes it expected,  and we avoid app crashes if returned value null by map it to empty string

data/repositories/vod/src/main/java/com/sanaa/vod/mapper/search/SearchActorMapper.kt
```diff
@@ -20,16 +29,30 @@ fun ActorSearchDto.toLocalDto(language: String): ActorLocalDto {
...
     )
 }
 
 internal fun getFullImageUrl(path: String?): String =
-    if (path.isNullOrBlank()) "" else "${TMDB_IMAGE_BASE_URL}$path"
\ No newline at end of file
+    if (path.isNullOrBlank()) "" else "${TMDB_IMAGE_BASE_URL}$path"
+
+fun mapGender(gender: Int?): Actor.Gender {
+    return if (gender == 1) Actor.Gender.FEMALE else Actor.Gender.MALE
```
Radwa85, line 57: this function will treat null as male we should handle that

PR #366, opened 2025-07-29
Refactor/ Fix Genres Localization

### LINKED ISSUE #355: issue:home: genres localization
when switch language, update the genres in home screen and other related screens (top rated, trending movies, trending tv shows)

AhmadKharfan:
Screenshot 2025-07-29 173258
Screenshot 2025-07-29 173321


ZainabAlnajjar (changes requested):
the problem was that the genres were not localized when switching the language while the app is still running, you should detect the language change and refresh genres
image

ElhamFadel (approved):

AhmadKharfan (commented):
@ZainabAlnajjar 
The problem I fixed here was that the data was only being fetched in one language in the first place
The issue you mentioned (genres not updating when switching the language while the app is still running) exists across all screens the language doesn’t actually update unless another request is sent to the API
This will require a separate PR to handle language change detection and trigger data refresh for all screens
So for now this PR focuses only on fixing the single language fetching problem

ZainabAlnajjar (approved):

PR #405, opened 2025-07-31
feature implement watching from remote data source

### LINKED ISSUE #402: implement watching data

raouf-kamel:
add fetch Watchlist Movies ,
fetch Watchlist Tv Shows,
add To Watchlist,


domain/vod/src/main/java/usecase/ManageMovieUseCase.kt
```diff
@@ -42,6 +43,33 @@ class ManageMovieUseCase(
...
+            body = watchlistInfo.copy(
+                mediaType = watchlistInfo.mediaType,
+                mediaId = watchlistInfo.mediaId,
+                watchlist = watchlistInfo.watchlist
+            )
+
+        )
+
+    }
+
```
ZainabAlnajjar: I believe we shouldn't pass the authorization and accountId from the usecase, these things are not the vod responsibility

if you need these informations, I suggest to inject what you need in the repository implementation and you it directly there

ZainabAlnajjar: also we need to filter by genre, so please add genreId to the functions parameters and do the filter in the datasource result

data/remote_data_source/vod/src/main/java/com/sanaa/vod/media/movie/MovieApiService.kt
```diff
@@ -86,4 +92,26 @@ interface MovieApiService {
...
+    @GET("account/{account_id}/watchlist/movies")
+    @Headers(
+        "Ignore-Language: true",
+        "accept: application/json"
+    )
+    suspend fun fetchWatchlistMovies(
+        @Header("Authorization") authorization: String,
+        @Path("account_id") accountId: String,
+        @Query("page") page: Int,
+        @Query("sort_by") sortBy: String = "created_at.asc",
```
AimanYosofiUlt, line 113: What is this for 🤔
` "created_at.asc"`

data/remote_data_source/vod/src/main/java/com/sanaa/vod/media/movie/MovieApiService.kt
```diff
@@ -86,4 +92,26 @@ interface MovieApiService {
...
+    ): PaginatedResponse<MovieDto>
+
+    @Headers(
+        "accept: application/json",
```
AimanYosofiUlt: We don't need to pass this. If we are actually going to pass it, it should be passed inside an  interceptor

data/remote_data_source/vod/src/main/java/com/sanaa/vod/media/movie/MovieApiService.kt
raouf-kamel: Sort by save to playlist or watch

PR #413, opened 2025-07-31
added wavyProgressIndicator when loading state

ElhamFadel:
https://github.com/user-attachments/assets/389b5ff7-d373-402c-8850-5ed1b4d8ed92


feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchViewModel.kt
```diff
@@ -341,12 +341,14 @@ class SearchViewModel(
     ): Flow<PagingData<R>> {
         return Pager(
             config = PagingConfig(
-                pageSize = PAGE_SIZE, enablePlaceholders = false
-            ), pagingSourceFactory = pagingSourceFactory
+                pageSize = PAGE_SIZE, 
+                enablePlaceholders = false,
+                prefetchDistance = 4
+            ),
```
haideraljawasim, line 336: Why would you set prefetchDistance to 4

feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchViewModel.kt
```diff
@@ -357,6 +359,10 @@ class SearchViewModel(
         updateState { it.copy(isLoading = false, noInternetConnection = false) }
     }
 
+    private fun setErrorState(error: String) {
+        updateState { it.copy(isLoading = false, error = error, noInternetConnection = false) }
+    }
+
```
ZainabAlnajjar: you added the function, but I don't see it is used anywhere

feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchViewModel.kt
ElhamFadel: just for enhance performance so this response to get 4 items before next request

feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchViewModel.kt
ElhamFadel: I was plan to handle error that happened but I saw we don't need it

AimanYosofiUlt (approved):

shahad-97 (approved):

PR #414, opened 2025-07-31
Refactor/ Fix Issue In HorizontalPager Home Screen

### LINKED ISSUE #412: prevent 2-page jumps and always reset to page 0 in HorizontalPager

AhmadKharfan:
## what’s fixed
- **auto-scroll jitter** – carousel no longer jumps two cards at a time.
- **wrap-around bug** – after the last card it always restarts at page 0.
- **dot click crash** – calling `animateScrollToPage()` from a coroutine removes the “suspend function” error.

## how it works now
1. there is a single `LaunchedEffect` tied to the pager state, so only one timer drives the auto-scroll.  
2. the next index is calculated with `(currentPage + 1) % pageCount`, ensuring a clean loop.  
3. dot clicks run `animateScrollToPage()` inside `rememberCoroutineScope().launch { … }`.

### DIFF
--- a/feature/home/presentation/src/main/java/com/sanaa/presentation/screen/homeScreen/section/PopularMediaSection.kt
+++ b/feature/home/presentation/src/main/java/com/sanaa/presentation/screen/homeScreen/section/PopularMediaSection.kt
@@ -11,17 +11,7 @@ import androidx.compose.animation.fadeOut
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.isSystemInDarkTheme
-import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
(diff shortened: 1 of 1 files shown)


feature/home/presentation/src/main/java/com/sanaa/presentation/screen/homeScreen/section/PopularMediaSection.kt
```diff
@@ -75,29 +68,55 @@ fun PopularMediaSection(
...
-            onDotClick = {})
+            onDotClick = { index ->
+                coroutineScope.launch {
+                    pagerState.animateScrollToPage(
+                        page = index,
+                        animationSpec = tween(durationMillis = 600, easing = FastOutSlowInEasing)
+                    )
+                }
+            }
+        )
```
ZainabAlnajjar, line 112: the dots are too small that is why it didn't has onClick event

ZainabAlnajjar (approved):

haideraljawasim (approved):

shahad-97 (approved):

feature/home/presentation/src/main/java/com/sanaa/presentation/screen/homeScreen/section/PopularMediaSection.kt
AhmadKharfan: If a user wants to jump back to the first item they can just tap the first dot in other words they don’t have to wait for the auto scroll cycle and that’s exactly how most carousel UIs work in popular apps
Anyway The dots are an optional

PR #444, opened 2025-08-01
Refactor/filter flow

### LINKED ISSUE #437: remove search filter
The current available endpoint doesn't support filtering the query, we are requesting many pages just to get few results in some cases when the filter is rare to match, therefore the user would wait for long time in loading state and that would give bad user experience, since we can't control the backend (endpoint), the app is better off without these filter for smoother user experience

Note: there is already a branch to refactor search pagination with filters, so we can leave this issue to that branch #428

EL-MANCY:
Remove all filter Flow


feature/search/presentation/src/main/java/com/sanaa/presentation/screen/SearchScreen.kt
```diff
@@ -106,8 +95,8 @@ fun SearchScreenContent(
         SearchSection(
             text = uiState.searchQuery,
             onTextChange = searchListener::onSearchQueryChanged,
-            onFilterClicked = { searchListener.onFilterClicked() },
-            isFilterButtonVisible = uiState.isFilterVisible(),
+            onFilterClicked = { /* No-op */ },
+            isFilterButtonVisible = false,
```
haideraljawasim: Remove this since they related to filter bottom sheet

haideraljawasim (approved):

feature/search/presentation/src/test/java/com/sanaa/presentation/screen/SearchViewModelTest.kt
```diff
@@ -371,25 +354,19 @@ class SearchViewModelTest {
     }
 
     @Test
-    fun `onTabSelected() should load actors when actor tap selected and hide filter button`() =
+    fun `onTabSelected() should load actors when actor tap selected`() =
```
ZainabAlnajjar, line 355: could you also fix the typo on tap and make it tab?

feature/search/presentation/src/test/java/com/sanaa/presentation/screen/SearchViewModelTest.kt
```diff
@@ -371,25 +354,19 @@ class SearchViewModelTest {
...
-                    uiState.value.searchQuery, page
-                )
-            } returns actors
+            coEvery { searchUseCase.searchActors(actorName, 1) } returns actors
 
             searchViewModel.onTabSelected(index)
 
             searchViewModel.state.test {
+                awaitItem()
                 awaitItem()
```
ZainabAlnajjar: I don't think you need 2 await

ZainabAlnajjar (approved):

PR #447, opened 2025-08-01
Refactor/api key interceptor

raouf-kamel:
refactor api key interceptor


shahad-97:
Nice work Raouf 
 I really like how you're using an `Interceptor` to inject the `api_key` and `session_id` dynamically  clean and flexible approach
 I have a few small suggestions that might help improve the code:
 - There's a bit of redundancy in building the URL consider using a single `urlBuilder` variable to keep things clean
 - If the API doesn’t require `session_id` as a header it’s probably best to just include it as a query parameter

PR #492, opened 2025-08-05
Feature/cache home data

### LINKED ISSUE #475: feature/ cache home screen data for 24 hours
Make Home screen cache all data to improve application UX, so when the user opens app, it won't spend sometime until showing the data. cache should expire after 24 hours.

ZainabAlnajjar:
# What have been done
- create local database for caching home screen sections content 
- applied relationship between tables to ensure that movies and tv shows won't be duplicated in memory
- ensure that data would be updated when changing language
- ensure every method is still functional with pagination
- clear cached content every 24 hours
- add test cases for all the effected functionality

Note: there is a problem with home screen viewmodel that shows `NetworkDisconnectionContact` when other data like genres are not fetched, this would be fixed along with other bugs in #493 

Photo of cached media after disabling the `NetworkDisconnectionContact` state on home screen content:
image


haideraljawasim (commented):
Create work Zainab!
I run the project but i couldn't see the cached data in Home screen when open the app and wifi off could you please check again or record a video cause maybe the problem only happen with me

data/repositories/vod/src/main/java/com/sanaa/vod/dataSource/local/cache/dto/CachedContentMetadataLocalDto.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.Index
+import androidx.room.PrimaryKey
+
+@Entity(
+    tableName = "cached_content_metadata",
+    indices = [Index(value = ["category"], unique = true)
+    ]
```
Radwa85, line 11: Please use a composite index on category and language instead of category only this will improve lookup performance and avoid full table scans . like that indices = [Index(value = ["category", "language"])]

ZainabAlnajjar:
@haideraljawasim the problem is in the viewmodel state not the caching (cuz we fetch other things like genres and when it fail, we got the network error screen),
please disable the "NetworkDisconnectionContact" screen content and you will see the cache

I am working on this bug in another branch to refactor home feature along with solving any bug in its state
image

data/repositories/vod/src/main/java/com/sanaa/vod/dataSource/local/cache/dto/CachedContentMetadataLocalDto.kt
ZainabAlnajjar: I made it only on category on purpose, so when the language change I won't be caching the not-needed language anymore (all related content info would be deleted, while the media table would be updated),
also because the movies/tvShows would be cached in one language based on the id, then replaced if the language changed, therefore I wont be storing un-needed language media (and won't need to add language parameter to the media to avoid getting the wrong language)

do you think caching media for both languages would be better than replacing it?

Radwa85: Thanks for the clarification ,your approach makes perfect sense if the app is meant to cache data in one language only i agree with you now

haideraljawasim (approved):

shahad-97 (approved):

PR #519, opened 2025-08-07
Refactor/home feature

### LINKED ISSUE #493: refactor/home feature
simplify the code of home feature and solve any bugs related to home feature

ZainabAlnajjar:
# What have been done:
- refactor code and unify variables and function names
- handle all error and network disconnection error for all screens
- handle loading state on all pagination section with loading indicator
- add snack bar and refresh button to all screens with localized error message using string provider
- fix bug in home loading state, and no network connection screen


feature/home/presentation/src/main/java/com/sanaa/presentation/components/RequestToLoginBottomSheet.kt
```diff
@@ -58,7 +58,7 @@ fun RequestToLoginBottomSheet(
                 Image(
                     painter = painterResource
                         (id = loginImageId),
-                    contentDescription = "pleas login light",
+                    contentDescription = "please login",
```
AhmadKharfan, line 61: we can also extract this to string resource file

AhmadKharfan (approved):

haideraljawasim (commented):
Create job Zainab,
But could you recheck on the screen cause here in your branch you only display Tv shows In popular and top rated section

shahad-97 (approved):

domain/vod/src/main/java/service/VodStringProvider.kt
```diff
@@ -0,0 +1,8 @@
+package service
+
+interface VodStringProvider {
+    val noInternetConnectionError: String
+    val somethingWentWrongError: String
+    val addToListFailed: String
+    val addToListSuccess: String
+}
```
ElhamFadel, line 8: Why you use ```interface``` then declare varbs with  ```val``` 

if this varbs changes in run time you can deal with it like function

ElhamFadel: Also name of interface not clear 
String for what ? is it messages or alerts

ZainabAlnajjar: they are strings that can change by language, I am providing them from repository implementation

Radwa85 (approved):

PR #538, opened 2025-08-08
Refactor/local nav controller provider

ElhamFadel:
Transfered LocalNavProvider to design system models


design_system/src/main/java/com/sanaa/designsystem/design_system/component/navigation/LocalNavControllerProvider.kt
```diff
@@ -0,0 +1,9 @@
+package com.sanaa.designsystem.design_system.component.navigation
+
+import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.navigation.NavHostController
+
+
+val LocalNavControllerProvider = staticCompositionLocalOf<NavHostController> {
```
Radwa85, line 7: i do not think it is true to but it into design system

ElhamFadel: where you think we can added it ?

haideraljawasim: I agree with Radawa its not related design system, and our modularization force us each feature should be into sprite module  and can't one depend directly on the other, also we can't add shared module  , so we don't have any other option only duplication 
@ElhamFadel

ElhamFadel: so we don't need make refactor here

PR #544, opened 2025-08-09
Feature/tv navigation

shahad-97:


tvapp/build.gradle.kts
```diff
@@ -0,0 +1,138 @@
...
+    implementation(projects.feature.search.presentation)
+    implementation(projects.feature.mediaDetails.api)
+    implementation(projects.feature.mediaDetails.presentation)
+    implementation(projects.feature.userProfile.api)
+    implementation(projects.feature.userProfile.presentation)
+    implementation(projects.feature.playlists.api)
+    implementation(projects.feature.playlists.presentation)
+    implementation(projects.feature.authentication.api)
+    implementation(projects.feature.authentication.presentation)
+    implementation(projects.preferences)
```
AhmadKharfan, line 133: These features are not needed in the TV module consider removing them even though you didn’t add them here

tvapp/src/main/java/com/sanaa/tvapp/TvApp.kt
```diff
@@ -0,0 +1,8 @@
+package com.sanaa.tvapp
+
+import android.app.Application
+import dagger.hilt.android.HiltAndroidApp
+
+@HiltAndroidApp
+class TvApp: Application() {
+}
```
AhmadKharfan, line 8: Application class for the TV module currently missing onCreate() for initialization
```suggestion
@HiltAndroidApp
class TVApp : Application() {

    override fun onCreate() {
        super.onCreate()
    }
}
```

AhmadKharfan (commented):
This navigation setup differs from the rest of the app.
We normally use type-safe route classes with PATTERN/ARG constants, dynamic start routes,
CompositionLocal for NavController, and Hilt ViewModel injection per destination.
Please align with the existing navigation pattern used in other modules for safety,
to avoid hardcoded values, ensure type-safety, maintain consistency across the app,
 and make future changes easier.

PR #564, opened 2025-08-11
Fix/rating icon

ElhamFadel:


feature/media_details/presentation/src/main/java/com/sanaa/presentation/shared_component/BottomContainer.kt
```diff
@@ -42,7 +44,10 @@ fun BottomContainer(
             PrimaryButton(
                 text = null,
                 onClick = onSetRateClicked,
-                icon = painterResource(R.drawable.icon_star_outlined)
+                icon = painterResource(
+                    if (isRateSelected) R.drawable.icon_star else R.drawable.icon_star_outlined
+                ),
+                iconTint = if (isRateSelected) Theme.colors.statusColors.yellowAccent else Theme.colors.onPrimary
```
ZainabAlnajjar: I just checked on the design to see the icon color, but I think the requirements has changed, we should remove the button if the user already rated this movie (no need to handle the stars state if there is rate found)
image

mostafa-n3ma:
can you please just add some description to your work what you have done here 
and make sure to update the branch

shahad-97 (approved):

PR #579, opened 2025-08-12
Refactor/profile feature

### LINKED ISSUE #570: Profile Screen Refactor
2

haideraljawasim:
# What have been done :
- Refactor User Profile base feature 
- Refactor all of MyAccountViewModel , MyAccountScreen and MyAccountUiState 
- Split the code into small components 
- Rename the packages to align with other features


feature/user_profile/presentation/src/main/java/com/sanaa/presentation/ProfileViewModel.kt
```diff
@@ -25,30 +25,38 @@ class MyProfileViewModel @Inject constructor(
...
-                mangeUserPreference.getTheme().collect { theme ->
-                    updateState { it.copy(isDarkTheme = theme == Theme.DARK) }
-                }
+            fetchTheme()
+            fetchContentRestriction()
+        }
+        updateState { it.copy(isReady = true) }
+    }
+
+    private fun fetchTheme() {
```
mostafa-n3ma, line 34: make this function suspend cause you already called it from a viewModel scope line 28

mostafa-n3ma (changes requested):
nice work may man ,
just make sure to update that fetch method I commented on earlier

feature/user_profile/presentation/src/main/java/com/sanaa/presentation/ProfileViewModel.kt
haideraljawasim: its don't need it to be suspended cause the main function called in another view model scope  ,  and it's not  a must to call only suspended functions inside view mode scope(Note I'm calling them here inside view model scope to set the ready true after both are executed)    , Last thing this it collecting flow and flow don't need to be suspended !! , so can you please explain me your idea so you felt the need to request changes for it

mostafa-n3ma (approved):

shahad-97 (approved):

raouf-kamel (approved):
good job

AimanYosofiUlt (approved):

PR #586, opened 2025-08-12
Fix/date parser

### LINKED ISSUE #583: fix: date parser
the app crash when the api return the date as empty string, also remove the default 1900-1-1 date when api return null date

ZainabAlnajjar:
# What have been done
- fix crash of returning null or empty date from api 
- all entities has non-nullable date now
- add function to parse the date safely or return default value of LocalDate(1, 1, 1) if date is invalid
- handle the above changes in ui


data/repositories/vod/src/main/java/com/sanaa/vod/repository/mapper/media/episodeMapper.kt
```diff
@@ -13,7 +13,7 @@ fun EpisodeDto.toEntity(): Episode {
         number = episodeNumber,
         imdbRating = voteAverage,
         durationMinutes = runtime,
-        releaseDate = airDate?.let { LocalDate.parse(it) } ?: LocalDate.parse("1970-01-01"),
+        releaseDate = getLocalDateOrDefault(airDate),
```
mostafa-n3ma, line 16: try : safeParseLocalDate(airDate)

data/repositories/vod/src/main/java/com/sanaa/vod/util/DateTimeUtils.kt
```diff
@@ -0,0 +1,19 @@
...
+import kotlinx.datetime.LocalDate
+
+object DateTimeUtils {
+    fun getCurrentTimeStamp(): Long {
+        return System.currentTimeMillis()
+    }
+
+    fun getLocalDateOrDefault(date: String?): LocalDate {
+        return try {
+            LocalDate.parse(date ?: "")
```
mostafa-n3ma, line 12: the try catch feels like you are handling the failure exception  without validating first

AhmadKharfan (approved):

data/repositories/vod/src/main/java/com/sanaa/vod/util/DateTimeUtils.kt
ZainabAlnajjar: it will throw exception either ways, didn't want to make more unnecessary checks

AimanYosofiUlt (approved):

PR #630, opened 2025-08-17
fix/remote-image-re-composition

ahmedfikry24:


AimanYosofiUlt (approved):
Thanks for your effort

image_viewer/src/main/java/com/sanaa/image_viewer/component/RemoteBlurredSensitiveImage.kt
```diff
@@ -62,77 +65,71 @@ fun RemoteBlurredSensitiveImage(
...
 
-    LaunchedEffect(imageUrl) {
-        requestState = RequestState.LOADING
+    val imageRequest = remember(isImageLoaded, contentThreshold) {
+        ImageRequest.Builder(context)
+            .data(imageUrl)
+            .isBlurEnable(
+                enabled = isBlurEnabled && isSensitive,
+                context = context,
+                blurValue = contentThreshold * 25 // blur value range is from 1 to 25
```
ZainabAlnajjar: I don't think the blur radius should be controlled by the threshold

we have blurRadius parameter, we can pass it here and set the parameter range values
this would leave the option for the developer to keep the blur radius the same even with different threshold

raouf-kamel (approved):
Nice work, fikry

ZainabAlnajjar (approved):

PR #696, opened 2025-08-21
Fix/guest login

EL-MANCY:


tvapp/src/main/java/com/sanaa/tvapp/presentation/screens/myAccount/MyAccountScreen.kt
```diff
@@ -75,8 +81,17 @@ fun MyAccountScreen(viewModel: MyAccountScreenViewModel = hiltViewModel()) {
...
+    val context = LocalContext.current
     val activity = view.context as? AppCompatActivity
 
+    val loginLauncher = rememberLauncherForActivityResult(
+        contract = ActivityResultContracts.StartActivityForResult()
+    ) { result ->
+        if (result.resultCode == Activity.RESULT_OK) {
+            activity?.recreate()
+        }
+    }
```
mostafa-n3ma, line 93: my dear abdo , we should set our own result codes , to know if the user logged in with session id or logged as guest 
check the auth launcher in the authentication module

mostafa-n3ma: or do you have another  way to distinguish that ?

EL-MANCY: Why do we need to do that 
Its only one condition when the user is back from a another screen to login

shahad-97 (approved):

raouf-kamel (approved):
Nice work

